PR #4, opened 2022-06-03
Revert "add: Include SCA info in semgrep-app findings output"

mmcqd:
Reverts returntocorp/semgrep-interfaces#3


brendongo (approved):

aryx:
There was no need to revert that. It was an optional field anyway

PR #6, opened 2022-06-08
Update to add pattern based id spec

ajbt200128:
Added the field `pattern_based_id` to Finding


brendongo (approved):
This looks good to me! Idk if @aryx we have to do anything special after modifying the interface or we can just update the pointer in semgrep repo and everything just works?

Semgrep_output_v0.atd
```diff
@@ -576,6 +576,7 @@ type finding = {
   index: int;
   commit_date: string;
   syntactic_id: string;
+  pattern_based_id: string;
```
aryx, line 579: I usually add a comment such as (* since semgrep 0.98 *)
and I usually make the field optional, since it's not present in older version, and sometimes we want
to remain backward compatible (especially in the code that reads the findings, such as the App,
which may get findings from older version of semgrep still).

aryx (approved):

aryx:
yes, just update the submodule in semgrep, run mypy which should tell you where you need to fix the code (or if the field is optional, you will need to find where to add the code yourself).

After for the server-side you'll have also to update a submodule. Nathan knows more on this.

aryx (changes requested):
Need to make the field optional I think.

ajbt200128:
@aryx also update the name to match_based_id?

PR #8, opened 2022-06-21
Add `deriving show` for OCaml type definitions

kopecs:
Adds a derive for `show` so you can print debug the types defined here.

### DIFF
--- a/semgrep_output_v0.atd
+++ b/semgrep_output_v0.atd
@@ -57,19 +57,20 @@
 (*****************************************************************************)
 
 (* escape hatch *)
-type raw_json <ocaml module="Yojson.Safe" t="t"> = abstract
+type raw_json <ocaml module="Yojson.Safe" t="t" attr="deriving show"> = abstract
 
 (*****************************************************************************)
 (* Versioning *)
 (*****************************************************************************)
 (* semantic versioning, see http://semver.org/spec/v2.0.0.html *)
-type semver = string (* e.g., "0.91.0" *)
+type semver <ocaml attr="deriving show"> = string (* e.g., "0.91.0" *)
 
 (*****************************************************************************)
 (* Location *)
 (*****************************************************************************)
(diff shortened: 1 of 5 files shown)


kopecs:
Is it intentional that autogenerated files are tracked by `git`?

kopecs:
Force pushing to add autogenerated files for now.

mjambon:
> Adds a derive for show so you can print debug the types defined here.

You should be able to do this by passing a flag to the `atdgen` command in the `dune` file, rather than annotating each type definition in the atd file.

Right now, there's in-between option for setting this in the atd file but only in one spot at the beginning of the file. Maybe that would be a good thing to add to atdgen.

mjambon:
> Is it intentional that autogenerated files are tracked in the git history? I would've figured these would've been ignored by git and we'd just generate them when the cli or core makefile ran a target requiring them.

The goal is to make these interfaces viewable by people who are not familiar with the code generators or don't want to deal with that. In general, this practice poses some problems (maintenance, security) but alternatives are more complicated to set up. To do things well, we'd need a source repo and a separate repo/tarball. The latter would contain the generated code, pushed automatically by a CI job.

mjambon (approved):

kopecs:
> You should be able to do this by passing a flag to the atdgen command in the dune file, rather than annotating each type definition in the atd file.

"Should" as in this is currently possible and I should change the commands we use, or as a normative statement?

aryx:
I would prefer to be able to set the default value in the file rather than in the dune file with special options to atdgen. It also generalizes better to other annotation-like features in other languages (e.g., python decorator, Java/Scala annotations).

mjambon:
> > You should be able to do this by passing a flag to the atdgen command in the dune file, rather than annotating each type definition in the atd file.
> 
> "Should" as in this is currently possible and I should change the commands we use, or as a normative statement?

"Should" as in "it's currently possible but I haven't tried it myself".

PR #9, opened 2022-07-05
Add taint traces to semgrep output

nmote:
This is a simplified version of semgrep-core's `Pattern_match.taint_trace` type. It makes a couple of simplifications:
* It omits the sink. The sink is already going to be the primary location of the error message, except for more complicated behavior with DeepSemgrep. We can add it in later if we need to.
* It simplifies the source. The `Call` case is only used for DeepSemgrep, and atdpy does not support recursive types so it would be difficult to represent here anyway. Instead, we can just provide a list of locations.

I believe that despite these simplifications, we will still be able to achieve our goals of providing taint trace information to end users and improving our ability to make PR/MR comments.

This is also meant to be extended to other kinds of dataflow traces (such as explaining constant propagation) without a breaking change.

### DIFF
--- a/semgrep_output_v0.atd
+++ b/semgrep_output_v0.atd
@@ -108,6 +108,24 @@ type core_match_extra
   <python decorator="dataclass(frozen=True)"> = {
   ?message: string option; (* rule.message (?) *)
   metavars: metavars;
+  ?dataflow_trace: dataflow_trace option;
+}
+
+(* Provides information about dataflow that leads to a finding. For now, just
+ * provides taint traces (i.e. information about the taint source and how the
+ * source reaches the sink). In the future, we may want to include more dataflow
+ * information. For example, we may want to explain how constant propagation
+ * leads to a finding. *)
+type dataflow_trace
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  ?taint_source: location list option;
+  (* Intermediate variables which are involved in the dataflow. For taint, this
+   * explains how the taint flows from the source to the sink. *)
+  ?intermediate_vars: location list option;
+  (* For now, omitting the taint sink, since it's redundant data except for
+   * certain cases that only appear with DeepSemgrep. We can add it in later
+   * without breaking consumers. *)
 }
 
 (* Name/value map of the matched metavariables.
@@ -477,6 +495,10 @@ type cli_match_extra <ocaml attr="deriving show"> = {
    * and seems to be used only when sending findings to the App.
    *)
   ?fixed_lines: string list option;
+
+  (* For now, present only for taint findings. May be extended to others later
+   * on. *)
+  ?dataflow_trace: dataflow_trace option;
 }
 
 type fix_regex <ocaml attr="deriving show"> = {
--- a/semgrep_output_v0.ts
+++ b/semgrep_output_v0.ts
@@ -35,6 +35,12 @@ export type CoreMatch = {
 export type CoreMatchExtra = {
   message?: string;
   metavars: Metavars;
+  dataflow_trace?: DataflowTrace;
+}
+
+export type DataflowTrace = {
+  taint_source?: Location[];
+  intermediate_vars?: Location[];
 }
 
 export type Metavars = [string, MetavarValue][]
@@ -205,6 +211,7 @@ export type CliMatchExtra = {
   dependency_match_only?: boolean;
   dependency_matches?: RawJson;
   fixed_lines?: string[];
+  dataflow_trace?: DataflowTrace;
 }
 
 export type FixRegex = {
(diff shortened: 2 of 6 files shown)


IagoAbal (approved):
Looks good to me! This is all CLI/App need right now, and I like that we keep it simple. Also, the "deep trace" is going to change anyways, I already realized that it isn't enough to capture all cases.

IagoAbal:
> Don't merge this yet, just soliciting feedback for now.

Just FYI you could mark a PR as "DRAFT" and then I couldn't merge by accident.

> * It simplifies the source. The `Call` case is only used for DeepSemgrep, and atdpy does not support recursive types so it would be difficult to represent here anyway. Instead, we can just provide a list of locations.

Agree with the simplification. But we'll have to do something about this one day. I thought that atd supported recursive types, maybe not yet for Python. 🤔

nmote:
Unfortunately I believe that if I mark a PR as a draft then it won't notify reviewers about it.

Right, I gathered from the message that ATD does support recursive types, but atdpy specifically does not. Maybe this is something we could address down the road.

I'll run this by App tomorrow and then merge if there is consensus. Thanks!

semgrep_output_v0.atd
```diff
@@ -108,6 +108,20 @@ type core_match_extra
   <python decorator="dataclass(frozen=True)"> = {
   ?message: string option; (* rule.message (?) *)
   metavars: metavars;
+  (* Optional: present only for taint rule matches *)
+  ?taint_trace: taint_trace option;
+}
+
+type taint_trace
```
nbrahms, line 115: Recommendation:

We may want to use this for other dataflow applications beyond taint. @kopecs had a suggestion to instead return an array of context information with arbitrary tags (perhaps with labels drawn from an enum).

In that case you might get something more like (in TS, b/c I'm not super familiar with ATD):

```typescript
type DataflowTag = "taint_source" | "taint_intermediate_var";
type DataflowContextItem = {
  tag: DataflowTag;
  location: Location;
}
type DataflowContext = array[DataflowContextItem];
```

nmote: As we discussed elsewhere, your proposal has the downside that adding additional variants later is a breaking change. If consumers don't handle a default case, we have to make sure to upgrade them first, adding complexity. It also loses some of the structure that the current proposal provides.

I believe we could accomplish the same thing by leaving the current structure as-is, but giving it a more generic name. If we want to add new kinds of dataflow context items, we can simply add them to the record type alongside `source` and `intermediate_vars`, potentially as an optional field. If `source` or `intermediate_vars` isn't needed, we can just use an empty list. I'd also want to rename `source` to `taint_source` in that case.

I believe this will allow us to extend this for other purposes later, maintain the structure, and avoid any breaking changes. What do you think?

semgrep_output_v0.atd
```diff
@@ -108,6 +108,20 @@ type core_match_extra
...
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  (* The source of taint. *)
+  source: location list;
+  (* Intermediate variables which are tainted *)
+  intermediate_vars: location list;
+  (* For now, omitting the sink, since it's redundant data except for certain
+   * cases that only appear with DeepSemgrep. We can add it in later without
+   * breaking backwards compatibility. *)
 }
```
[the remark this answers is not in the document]
nmote: The producer can just add new fields without introducing breaking changes for consumers. Do you have something else in mind?

semgrep_output_v0.atd
nbrahms: sure sgtm

semgrep_output_v0.atd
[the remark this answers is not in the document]
nbrahms: sgtm

nbrahms:
> Unfortunately I believe that if I mark a PR as a draft then it won't notify reviewers about it.

You can also leave a "changes requested" review on your own PR to make sure others don't merge it, I believe

semgrep_output_v0.ts
```diff
@@ -35,6 +35,12 @@ export type CoreMatch = {
 export type CoreMatchExtra = {
   message?: string;
   metavars: Metavars;
+  dataflow_trace?: DataflowTrace;
+}
+
+export type DataflowTrace = {
+  taint_source?: Location[];
```
zyannes, line 42: Are the locations guaranteed to be in order of evaluation?

zyannes (approved):

semgrep_output_v0.ts
IagoAbal: For now we don't plan on making such guarantees, unless it provides concrete benefits to "consumers".

nmote:
Thanks everyone!

PR #11, opened 2022-07-14
Move rule schema to semgrep_interfaces

kopecs:
This eliminates the need for separate PRs to update the version of this
file in Semgrep and Semgrep App.


rule_schema.yaml
```diff
@@ -0,0 +1,661 @@
+$id: https://raw.githubusercontent.com/returntocorp/semgrep/develop/cli/src/semgrep/rule_schema.yaml
+$schema: http://json-schema.org/draft-07/schema#
+$comment: "!!if you modify this file, you probably also need to update the one in semgrep-app:server/semgrep_app/rule_schema.yaml!!"
```
aryx, line 3: this comment is still valid?

aryx (approved):

rule_schema.yaml
kopecs: Until I PR to use this via the submodule it will be. I can just update it now though.

chmccreery (approved):
Thanks for doing this!

PR #13, opened 2022-07-20
Improve dataflow trace output

nmote:
This makes a breaking change to the dataflow_trace type, but since it is
not being consumed in production yet, the consensus is that this is
acceptable.

- Combine the multiple taint source tokens into one range spanning the
  entire taint source.
- Include `content` for taint source and intermediate vars.

### DIFF
--- a/semgrep_output_v0.atd
+++ b/semgrep_output_v0.atd
@@ -108,7 +108,7 @@ type core_match_extra
   <python decorator="dataclass(frozen=True)"> = {
   ?message: string option; (* rule.message (?) *)
   metavars: metavars;
-  ?dataflow_trace: dataflow_trace option;
+  ?dataflow_trace: core_match_dataflow_trace option;
 }
 
 (* Provides information about dataflow that leads to a finding. For now, just
@@ -116,16 +116,24 @@ type core_match_extra
  * source reaches the sink). In the future, we may want to include more dataflow
  * information. For example, we may want to explain how constant propagation
  * leads to a finding. *)
-type dataflow_trace
+type core_match_dataflow_trace
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True)"> = {
-  ?taint_source: location list option;
+  ?taint_source: location option;
   (* Intermediate variables which are involved in the dataflow. For taint, this
    * explains how the taint flows from the source to the sink. *)
-  ?intermediate_vars: location list option;
+  ?intermediate_vars: core_match_intermediate_var list option;
   (* For now, omitting the taint sink, since it's redundant data except for
    * certain cases that only appear with DeepSemgrep. We can add it in later
    * without breaking consumers. *)
+}
+
+type core_match_intermediate_var
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  location: location;
+  (* Making this its own type rather than inlining `location` because Iago has
+   * plans to extend this with more information *)
 }
 
 (* Name/value map of the matched metavariables.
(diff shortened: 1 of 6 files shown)


semgrep_output_v0.atd
```diff
@@ -119,15 +119,31 @@ type core_match_extra
...
   (* For now, omitting the taint sink, since it's redundant data except for
    * certain cases that only appear with DeepSemgrep. We can add it in later
    * without breaking consumers. *)
 }
 
+type taint_source
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  location: location;
+  abstract_content: string;
```
aryx, line 135: Why do we need those abstract_content?
Those abstract_content were hacks used a long time ago when the rule engine was split between OCaml and Python.
I'm not even sure we need them anymore in metavar_value.

semgrep_output_v0.atd
```diff
@@ -119,15 +119,31 @@ type core_match_extra
 type dataflow_trace
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True)"> = {
-  ?taint_source: location list option;
+  ?taint_source: taint_source option;
```
aryx, line 122: why taint_source is an option?

semgrep_output_v0.atd
nmote: @ajbt200128 specifically requested it to make the taint trace display easier in the App

semgrep_output_v0.atd
nmote: Based on feedback on the original PR, https://github.com/returntocorp/semgrep-interfaces/pull/9, I named this type `dataflow_trace` instead of `taint_trace` with the idea that it could be used to explain other kinds of dataflow that Semgrep performs. For example, we could explain how constant propagation leads to a finding. In that case it would not be explaining a taint mode finding, so there wouldn't be a taint source to include.

IagoAbal: We could also have kept it as an optional `taint_trace` and then add another optional `const_prop_trace` ?

semgrep_output_v0.atd
```diff
@@ -119,15 +119,31 @@ type core_match_extra
...
+  location: location;
+  abstract_content: string;
+}
+
+type intermediate_var
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  location: location;
+  abstract_content: string;
+  (* This happens to be the same as taint_source for the moment, but Iago has
```
IagoAbal, line 143: Does ATD support recursive data types? Because my extension is mostly for DeepSemgrep (it won't affect plain Semgrep) and it will require recursive data types.

nmote: It didn't let me use recursive types in an initial version of this datatype. My recollection is that ATD itself does, but atdpy does not. I'm not sure whether it just hasn't been implemented or whether it's due to some limitation of mypy.

semgrep_output_v0.atd
nmote: I think that would also have been a reasonable approach.

nmote:
I have updated this to use `content` instead of `abstract_content`. `content` is read from file in the CLI based on the locations provided by core, so I had to split out the types into cli_match and core_match variations.

aryx (approved):

semgrep_output_v0.atd
```diff
@@ -452,6 +460,38 @@ type position_bis <ocaml attr="deriving show"> = {
...
+}
+
+type cli_match_intermediate_var
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  location: location;
+  (* Unlike abstract_content, this is the actual text read from the
+   * corresponding source file *)
+  content: string;
+  (* This type happens to be the same as cli_match_taint_source for now, but
```
IagoAbal, line 491: The extension would be to have a call trace here, but I don't think these traces support "deep call traces" right now anyways no?

PR #16, opened 2022-07-27
Remove unique_id from semgrep_output_v0.atd

aryx:
This field was useful back in the day when the semgrep
engine was split between semgrep-core and semgrep-cli.
Now all the logic is in semgrep-core so there is no
need anymore for this unique_id field

test plan:
used in semgrep and everything seems to work fine

### DIFF
--- a/semgrep_output_v0.atd
+++ b/semgrep_output_v0.atd
@@ -157,27 +157,6 @@ type metavar_value
   end <ocaml name="end_">: position;
   abstract_content: string; (* value? *)
   ?propagated_value: svalue_value option;
-
-  (* The unique ID of a metavariable indicates the kind object that was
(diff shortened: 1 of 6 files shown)


aryx:
I will remove abstract_content is another diff.

nmote:
It looks like this is included in the CLI output, is that correct? If so, have you checked to make sure that downstream consumers are not relying on it?

nmote (approved):
Looks like @brendongo approved https://github.com/returntocorp/semgrep/pull/5820, so that suggests to me that the App is not in fact relying on this info.

aryx:
Yes I did before a 'git grep unique_id' in semgrep-app and there was no match (I should have put that in my test plan).

PR #17, opened 2022-07-27
Update comments for core parse errors

kopecs:
This is useful for new parse rate metrics, to prevent accidental
breakage if we add new core error kinds.

Relates to https://github.com/returntocorp/semgrep/pull/5815

### DIFF
--- a/semgrep_output_v0.atd
+++ b/semgrep_output_v0.atd
@@ -229,7 +229,9 @@ type core_error
 type core_error_kind
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True, order=True)"> = [
-  (* File parsing related errors *)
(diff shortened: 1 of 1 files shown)


brandonspark (approved):
let's go to mars

semgrep_output_v0.atd
```diff
@@ -229,7 +229,9 @@ type core_error
 type core_error_kind
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True, order=True)"> = [
-  (* File parsing related errors *)
+  (* File parsing related errors; if you add a target parse error then
```
aryx, line 232: I usually add a coupling: tag to the comment.
(* File parsing related errors.
 * coupling: if you do this ... you probably need to modify that ...
 *)

PR #18, opened 2022-07-28
feat(swift): added modifiers for Swift

brandonspark:
Swift has some keyword attributes which are not currently in the generic AST, namely `Throws` and `Rethrows`. I decided to just put it in here.

Let me know if that's not something that I should do.


nmote (approved):
Seems reasonable to me

aryx:
ast_generic_v0.atd is not really used for now so there is no need yet for those PRs to keep it sync with AST_generic.ml

PR #21, opened 2022-08-11
Modify rule_schema for templated rules

emjin:
For https://github.com/returntocorp/semgrep/pull/5907


aryx:
Closing for now, in favor of jsonnet experiment. We can still resume the PR if we want later.

PR #23, opened 2022-08-12
feat!: new rule syntax

kopecs:
TODO: longer description

This should probably be top level `oneOf`'d with the current schema; we support
backwards compatibility in the core PR

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -762,11 +762,11 @@ properties:
             properties:
               message: false
               severity: false
-              join: false
               pattern-sinks: false
               pattern-sources: false
               pattern-propagators: false
               pattern-sanitizers: false
+              join: false
         - if:
             properties:
               mode:
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -6,7 +6,7 @@ $schema: http://json-schema.org/draft-07/schema#
 #!!If you add new syntax to this file, you probably need to add some EXPERIMENTAL
 #  comment before!!
 $defs:
-  general-pattern-content:
+  new-pattern:
     title: "Return finding where code matches against the following pattern"
     oneOf:
     - type: string
@@ -145,11 +145,11 @@ $defs:
         - $ref: "#/$defs/pattern"
         - $ref: "#/$defs/pattern-regex"
         - $ref: "#/$defs/pattern-not-regex"
-        - $ref: "#/$defs/pattern-where-python"
         - $ref: "#/$defs/metavariable-analysis"
         - $ref: "#/$defs/metavariable-regex"
         - $ref: "#/$defs/metavariable-pattern"
         - $ref: "#/$defs/metavariable-comparison"
+        - $ref: "#/$defs/pattern-where-python"
   pattern-either-content:
     type: array
     title: "Return finding where any of the nested conditions are true"
@@ -217,6 +217,7 @@ $defs:
                       - patterns
                   - required:
                       - pattern-either
+  # EXPERIMENTAL
   r2c-internal-project-depends-on-content:
     type: object
     title: "One or more dependencies that the project contains in a lock file"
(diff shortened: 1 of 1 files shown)


emjin (approved):

aryx (changes requested):
let's hold we release semgrep 1.0 first.

brandonspark:
@aryx revisit?

rule_schema_v1.yaml
```diff
@@ -6,17 +6,103 @@ $schema: http://json-schema.org/draft-07/schema#
...
 $defs:
+  new-pattern:
+    title: "Return finding where code matches against the following pattern"
+    oneOf:
+    - type: string
+    - type: object
+      oneOf:
+        - required: [ pattern ]
+        - required: [ regex ]
+        - required: [ and ]
```
IagoAbal, line 17: I thought there was consensus in renaming `and` as `all` and `or` as `any`.

rule_schema_v1.yaml
```diff
@@ -723,14 +808,7 @@ properties:
               extract: false
               dest-language: false
               reduce: false
-              patterns: false
```
IagoAbal, line 726: Why is all this removed?

rule_schema_v1.yaml
```diff
@@ -6,17 +6,103 @@ $schema: http://json-schema.org/draft-07/schema#
...
     - type: string
-    # This is scuffed, but it seems that `allOf` does not short
-    # circuit.  This schema can potentially run on things which are
-    # not objects (such as lists), and if that happens, the validator
-    # will crash when it checks the keys of the object.  I would
-    # normally just use `allOf`, but since it doensn't short circuit,
-    # I had to instead make it this weird `if-then-else` thing to
-    # properly skip the keys check on a non-object.
+    # This is scuffed, but it seems that `allOf` does not short circuit.
+    # This schema can potentially run on things which are not objects (such as lists), and if that happens,
```
aryx, line 102: please limit sentences to 80 columns.

rule_schema_v1.yaml
```diff
@@ -637,7 +723,8 @@ properties:
         # EXPERIMENTAL
         r2c-internal-project-depends-on:
           $ref: "#/$defs/r2c-internal-project-depends-on-content"
-        # EXPERIMENTAL
```
aryx, line 640: keep EXPERIMENTAL comment
and also add it before extract.

rule_schema_v1.yaml
aryx: same question.

aryx (approved):

rule_schema_v1.yaml
brandonspark: was there consensus? i recall a discussion, but i dont feel qualified to make the call either way at this stage

i would be inclined to ship it with this syntax for now, and then test this with SR over the course of the next few weeks (i think most of the people weighing in were not security researchers). is that OK?

aryx: I recall bence had a pretty strong argument.
Note that security researchers do not have the final words on semgrep syntax. While our main target right now are AppSec people for semgrep, our long-term goal for semgrep is that every developers use Semgrep to enforce coding rules, finding bugs, enforcing security guardrails, etc. Therefore we should aspire to not only please security researchers but more generally every developers (hard to do, I know).

brandonspark: what i mean is that they have the most experience writing rules. i would be inclined to test it both ways, but they are professional users of the tool, so I would be inclined to see what they think

rule_schema_v1.yaml
brandonspark: think it was a weird rebasing thing

rule_schema_v1.yaml
brandonspark: anyways, ill make the change, but i am interested in what users think

PR #24, opened 2022-08-18
feat: Typed Supply Chain Info

mmcqd:
This PR defines a typed interface for supply chain findings.


aryx (approved):
A few comments here and there could help, but otherwise looks good.

PR #32, opened 2022-10-06
feat: dot operators

brandonspark:
## What:
Added a new `dot_operator` which will help us differentiate different kinds of DotAccesses.

## Why:
We want to be able to match in a way that is different for different kinds of dot accesses, such as `x.y` and `x?.y` in JS, or some more examples for Kotlin.

## How:
Added a new `dot_operator` type, which is extensible and may be used in the future to add other varieties of dot accesses.

### DIFF
--- a/ast_generic_v0.atd
+++ b/ast_generic_v0.atd
@@ -242,7 +242,7 @@ type expr = [
    * In the last case it should be rewritten as a (N IdQualified) with a
    * qualifier though.
   *)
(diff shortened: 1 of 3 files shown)


aryx:
you don't need to update that. It's not really used. You can just add some raise Todo in the code in semgrep.

brandonspark:
unnecessary now that we are no longer changing `dot_operator`

PR #34, opened 2022-11-02
Switch to v1

aryx:
This is to prepare for the release of Semgrep 1.0.0


aryx:
@zyannes this will probably have an impact on the app when you update to the latest version.

PR #36, opened 2022-11-07
feat(schema): update rule schema for syntax extensions

brandonspark:
## What:
This is meant to go along with https://github.com/returntocorp/semgrep/pull/6474. We update the rule schema here to match the corresponding syntax changes.

## Why:
Make SR lives easier!

## How:
Just updated the schema, in hopefully the right way.

## Test plan:
I can't add a test in this repo, but this file parses properly:
```
rules:
- id: rule-extensions 
  patterns:
  - pattern-not: "A"
  - pattern-not: 
      pattern: "A"
  - pattern-inside:
      pattern-regex: "B"
  - pattern-not-inside:
      pattern-not: "C"
  message: This rule is just meant to illustrate that we can parse "extensions" to old-grammar rules. 
  severity: WARNING
  languages: [python]
```

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -416,6 +416,7 @@ $defs:
     type: object
     properties:
       pattern-inside:
+        title: Return findings only from within snippets Semgrep pattern matches
         $ref: "#/$defs/general-pattern-content"
     required:
       - pattern-inside
@@ -424,6 +425,7 @@ $defs:
     type: object
     properties:
       pattern-not-inside:
+        title: Do not return findings from within snippets Semgrep pattern matches
         $ref: "#/$defs/general-pattern-content"
     required:
       - pattern-not-inside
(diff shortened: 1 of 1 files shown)


rule_schema_v1.yaml
```diff
@@ -2,6 +2,34 @@ $id: https://raw.githubusercontent.com/returntocorp/semgrep-interfaces/main/rule
 $schema: http://json-schema.org/draft-07/schema#
 $comment: "!!If you modify this file, you need to update the submodule in returntocorp/semgrep and returntocorp/semgrep-app!!"
 $defs:
+  general-pattern-content:
```
aryx, line 5: Could we use this new general-pattern-content at more places?

rule_schema_v1.yaml
```diff
@@ -388,26 +416,23 @@ $defs:
...
-        type: string
+        $ref: "#/$defs/general-pattern-content"
     required:
       - pattern-not-inside
     additionalProperties: false
   pattern-not:
     type: object
     properties:
       pattern-not:
-        title: Do not return finding where Semgrep pattern matches exactly
```
aryx, line 409: Why removing the title: ? Looks like we're losing useful comments.

aryx (approved):

PR #43, opened 2022-11-29
feat!: add memory usage

brandonspark:
Added an `ocaml_max_heap_words` field to the JSON output which denotes the maximum amount of words used by the OCaml process during execution. This will be useful for Metabase scanning, so we can plot this against the repository size and get memory benchmarks over time.

### DIFF
--- a/ast_generic_v1_j.ml
+++ b/ast_generic_v1_j.ml
@@ -1458,15 +1458,12 @@ let write_special = (
         ) ob x;
         Bi_outbuf.add_char ob ']'
       | `Require -> Bi_outbuf.add_string ob "\"Require\""
-<<<<<<< HEAD
       | `OtherSpecial x ->
         Bi_outbuf.add_string ob "[\"OtherSpecial\",";
         (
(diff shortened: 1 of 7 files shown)


aryx (approved):

aryx (commented):
Probably better to rename to max_memory_bytes
to be more independent of OCaml. This is an interface file, we're not supposed
to expose deep internals to the user.

brandonspark:
> Probably better to rename to max_memory_bytes to be more independent of OCaml. This is an interface file, we're not supposed to expose deep internals to the user.

I made it specific because I didn't want people not being aware of what this value meant. Is `max_memory_bytes` supposed to be for the OCaml or Python side? Also, the number I have here isn't in bytes, it's in the words of memory in the OCaml heap. Given that, I thought it would be better to be more _precise_ about what this number means, in case anyone looking at the output chooses to consume it. Thoughts?

aryx:
soon there will be no python, and bytes is less ambiguous than words in my opinion.

PR #44, opened 2023-01-04
Require rule key, and specific languages, plus some documentation

ajbt200128:
* Made sure the rules key is now required
* Added some documentation for metavariable-* keys
* Required the languages field be only one we support

I think the required languages field could be done better, such as moving [lang.json](https://github.com/returntocorp/semgrep-app/blob/develop/frontend/src/utils/lang.json) here and autogenerating it, but also that may be overcomplicating something that changes rarely.

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -217,9 +217,7 @@ $defs:
             id:
               $ref: "#/$defs/id"
             languages:
-              type: array
-              items:
-                type: string
+              $ref: "#/$defs/languages"
             pattern:
               type: string
             patterns:
@@ -252,11 +250,14 @@ $defs:
     properties:
       metavariable-analysis:
         type: object
+        title: Inspect a metavariable with a given analyzer
         properties:
           analyzer:
             type: string
+            title: Analyzer to use
           metavariable:
             type: string
+            title: Metavariable to analyze
           options:
             type: object
         required:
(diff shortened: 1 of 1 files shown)


rule_schema_v1.yaml
```diff
@@ -473,7 +482,55 @@ $defs:
   id:
     title: Rule ID to attach to findings
     type: string
-type: object
+  languages:
+    title: Languages this pattern should run on
+    type: array
+    items:
+      enum:
```
aryx, line 489: Is there a way to list an enum (so that some tools can leverage the info and provide a popup list in the playground)
but also accept any string? I'm worried we get some errors when we update lang.json and forget about updating this file.

Also can you add a big comment like "coupling: if you modify this, you first need to modify lang.json in the semgrep-langs repository"

aryx (approved):

rule_schema_v1.yaml
ajbt200128: Yea I can do both, I didn't realize there was a semgrep-langs :O is there a reason we don't include the repo in the app? [Here](https://github.com/returntocorp/semgrep-app/blob/develop/frontend/src/utils/lang.json) it looks like we have an outdated version that's just been copied in.

aryx: I dunno. They should use the semgrep-langs submodule (we do in semgrep).

PR #50, opened 2023-01-20
Add errors specifically for oom and timeout during preprocessing

emjin:
This will help us keep track of which errors are raised by pro engine users that prevent pro analysis from happening.

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -258,9 +258,6 @@ type core_severity
   <python decorator="dataclass(frozen=True)"> = [
   | Error <json name="error">
   | Warning <json name="warning">
-  (* For the timeout/OOM during preprocessing errors, where it is not
-     an error that crashes the program but should be shown to the user *)
-  | Alert <json name="alert">
 ] <ocaml repr="classic">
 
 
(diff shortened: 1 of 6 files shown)


semgrep_output_v1.atd
```diff
@@ -245,6 +245,9 @@ type core_error_kind
   | FatalError (* missing file, OCaml errors, etc. *) <json name="Fatal error">
   | Timeout <json name="Timeout">
   | OutOfMemory <json name="Out of memory">
+  (* pro-engine specific *)
+  | TimeoutDuringPreprocessing <json name="Timeout during preprocessing">
```
aryx, line 249: What is preprocessing in semgrep-pro? When we build the SASTs and SIGs for every files?

semgrep_output_v1.atd
```diff
@@ -255,6 +258,9 @@ type core_severity
   <python decorator="dataclass(frozen=True)"> = [
   | Error <json name="error">
   | Warning <json name="warning">
+  (* For the timeout/OOM during preprocessing errors, where it is not
```
aryx, line 261: Not a fan, but if you think it's sufficiently enough different than Error, why not.

aryx (approved):

semgrep_output_v1.atd
emjin: I ended up killing it and just having a special case

PR #51, opened 2023-01-24
Update generated TypeScript code to play well with tslint

mjambon:
Uses https://github.com/ahrefs/atd/pull/328


aryx:
and by playing well you mean just disabling it :)

PR #54, opened 2023-01-26
chore: label pro matches with engine kind

brandonspark:
in fairness emma did warn me

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -102,6 +102,13 @@ type rule_id
 (*****************************************************************************)
 (* EXPERIMENTAL: Do not rely on those internal types, they might disappear *)
 
+type engine_kind
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = [
+  | OSSMatch
+  | ProMatch
+] 
+
 type core_match
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True)"> = {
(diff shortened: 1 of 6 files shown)


aryx:
warn you about what?

semgrep_output_v1.atd
```diff
@@ -102,6 +102,13 @@ type rule_id
 (*****************************************************************************)
 (* EXPERIMENTAL: Do not rely on those internal types, they might disappear *)
 
+type engine_kind
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = [
+  | OSSMatch
```
aryx, line 108: if the type is called engine_kind, I would not add the Match suffix then
but OssEngine or ProEngine.

aryx:
And it's usually better to ask for reviewers on changes to semgrep-interfaces.

semgrep_output_v1.atd
```diff
@@ -586,6 +595,8 @@ type cli_match_extra <ocaml attr="deriving show"> = {
   (* EXPERIMENTAL: For now, present only for taint findings. May be extended to others
    * later on. *)
   ?dataflow_trace: cli_match_dataflow_trace option;
+
+  engine_kind: engine_kind;
```
aryx, line 599: This part should have a question mark before! You might break consumer of this!
See the toplevel comment in this file:
```
 * This file has the _v1 suffix to explicitely represent the
 * version of this JSON format. If you need to extend this file, please
 * be careful because you may break consumers of this format (e.g., the
 * Semgrep playground or Semgrep App backend or external users of this JSON). See
 * https://atd.readthedocs.io/en/latest/atdgen-tutorial.html?highlight=json#smooth-protocol-upgrades
 * for more information on how to smoothly extend the types in this file.
```

PR #55, opened 2023-01-26
chore: make engine_kind optional

brandonspark:


semgrep_output_v1.atd
```diff
@@ -128,7 +128,7 @@ type core_match_extra
    * finding. *)
   ?rendered_fix: string option;
 
-  engine_kind: engine_kind;
+  ?engine_kind: engine_kind option;
```
aryx, line 131: Here it's fine, you don't need the ?, because it's the interface between semgrep-core and semgrep CLI.
It's not public, nobody external should rely on it, so we can do whatever we want.

semgrep_output_v1.atd
```diff
@@ -600,7 +600,7 @@ type cli_match_extra <ocaml attr="deriving show"> = {
    * later on. *)
   ?dataflow_trace: cli_match_dataflow_trace option;
 
-  engine_kind: engine_kind;
+  ?engine_kind: engine_kind option;
```
aryx, line 603: here we need the ?, it's a public interface, it's the official semgrep output.

semgrep_output_v1.atd
brandonspark: In that case, could we make this just `engine_kind: engine_kind`? I believe that `semgrep-core` should only be able to inject into the `Some` case here, by construction.

semgrep_output_v1.atd
brandonspark: ah I see, I didn't realize that about the `?` (I thought they just went together always)

semgrep_output_v1.atd
aryx: yes, you can do engine: engine_kind.

semgrep_output_v1.atd
aryx: You can read this that I beleive martin wrote a long time ago:
https://atd.readthedocs.io/en/latest/atdgen-tutorial.html?highlight=json#smooth-protocol-upgrades

aryx: But basically by putting ?engine_kind, we will generate code that will accept the absence of engine_kind,
and so for example in the App, which use code generated from semgrep_output_v1.atd to read the output of semgrep,
they can read an engine_kind if it's there but also accept output where there is no engine_kind (such as with older version of semgrep, that we still want to run sometimes from the playground or also because customer may use older versions).

aryx (approved):

aryx:
why? it's needed for customers?

brandonspark:
@aryx the previous iteration of this would not be backwards-compatible right? i just wanted to prevent that possibility

aryx:
Oh yes, sorry, I was thinking of something else. Yes!

PR #58, opened 2023-02-04
feat!: schema for new taint

brandonspark:
Updated the rule schema for the new taint syntax.

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -32,6 +32,33 @@ $defs:
         type: string
       to:
         type: string
+  new-taint-content:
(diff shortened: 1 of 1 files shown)


aryx (approved):

rule_schema_v1.yaml
```diff
@@ -817,8 +821,16 @@ properties:
               - message
               - languages
               - severity
-              - pattern-sinks
-              - pattern-sources
+            if:
+              required:
```
aryx, line 825: "if required taint then required taint" ... I have no idea what this does ... jsonschema ...

aryx (approved):

PR #59, opened 2023-02-06
chore: add engine requested and rules by engine to output

brandonspark:
## What:
This PR adds in the engine requested by the Semgrep CLI, as well as a categorization of each rule run by which engine it was actually run with, to both the Core and CLI outputs.

This is important to consume to be transparent to users, since we may run rules with a different engine than what they actually requested. It is important to keep this information around for that reason.


semgrep_output_v1.atd
```diff
@@ -89,14 +89,21 @@ type location
...
-(* Rule id *)
+(* Rules *)
 (*****************************************************************************)
 
 (* e.g., "javascript.security.do-not-use-eval" *)
 type rule_id
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True)"> = string
 
+type rule
```
aryx, line 100: Please do not use rule. Rule means something special in Semgrep (see Rule.ml) where
we have a rule id, a message, a formula, etc.
I'd rather use rule_id_and_engine_kind

semgrep_output_v1.atd
```diff
@@ -441,6 +448,13 @@ type core_match_results
   stats: core_stats;
   (* LATER: rename timing *)
   ?time: core_timing option;
+
+  (* These rules, classified by engine used, will let us be transparent in the CLI
+     output over what rules were run with what.
+   *)
+  rules: rule list;
```
aryx, line 455: maybe an assoc from engine_kind to set of rules would be better than a flat list of rule_id_and_engine_kind?

Can the same  rule be run both with OSS and Pro?

aryx (changes requested):

semgrep_output_v1.atd
brandonspark: I thought about an `(engine_kind * rule_id list) list`, but I was afraid of the fact that, statically, you couldn't ensure that `engine_kind` appeared only once in the `assoc`. I'm fairly sure that a rule can only be run with one engine at a time, and I don't foresee that changing, so I went with the flat list. I can change it though, if you think it's better.

semgrep_output_v1.atd
```diff
@@ -632,6 +646,10 @@ type cli_output_extra <ocaml attr="deriving show"> = {
      *)
     (* EXPERIMENTAL: since semgrep 0.109? *)
     ?explanations: matching_explanation list option;
+
+    ?rules: rule list option;
```
aryx, line 650: you can add a since semgrep xxx comment before.

semgrep_output_v1.atd
aryx: In theory we could use (rule_id * engine_kind) list <json repr = "object">
which would make clearer that rule_id is the key and has one engine_kind,
but I guess rule_id_with_engine_kind list is fine too, but could be nice to add in comment
the invariant that's it's really a Map from rule_id to one engine_kind.

brandonspark: OK, renamed

semgrep_output_v1.atd
```diff
@@ -89,14 +89,21 @@ type location
 }
 
 (*****************************************************************************)
-(* Rule id *)
```
aryx, line 92: I would have kept Rule id for the section comment
We're not defining rules here.

semgrep_output_v1.atd
```diff
@@ -89,14 +89,21 @@ type location
...
 
 (* e.g., "javascript.security.do-not-use-eval" *)
 type rule_id
   <ocaml attr="deriving show">
   <python decorator="dataclass(frozen=True)"> = string
 
+type rule_id_and_engine_kind
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  rule_id: rule_id;
```
aryx, line 103: You can use a pair instead of a record. it's a bit ugly to see rule_id: rule_id; engine_kind: engine_kind;

aryx (approved):

semgrep_output_v1.atd
```diff
@@ -632,6 +647,10 @@ type cli_output_extra <ocaml attr="deriving show"> = {
      *)
     (* EXPERIMENTAL: since semgrep 0.109? *)
     ?explanations: matching_explanation list option;
+
+    (* since: 1.10.0 *)
```
aryx, line 651: Maybe you can add an EXPERIMENTAL, so people know those fields could change and
they should not rely on it.

aryx (approved):

PR #60, opened 2023-02-23
feat: add 2 new hashes to finding type

chmccreery:
The syntactic_id and match_based_id continue to not be exactly what we need 🙁 So, we need to start passing two new hashes: a pattern_hash (to replace match_based_id) and a code_hash (to replace syntactic_id).

These two new types will hash *only* the things we cannot pass in plaintext, i.e. information that requires code access. Everything else, which we already pass in plaintext (file path, rule name, index) will be kept separate from the hashes. This way, if in the future we want to count two findings as the same even if a user re-names a file, we have the information we need to do so.


dev.opam
```diff
@@ -10,4 +10,5 @@ synopsis: "OCaml development dependencies"
 depends: [
   "atdpy" {>= "2.10.0"}
   "atdts" {>= "2.10.0"}
+  "atdgen" {>= "2.10.0}
```
brandonspark, line 13: should there be a close quote here?

semgrep_output_v1.atd
```diff
@@ -787,6 +787,13 @@ type api_scans_findings <ocaml attr="deriving show"> = {
    cai_ids: string list;
 }
 
+type finding_hashes <ocaml attr="deriving show"> = {
+  start_line: string;
```
brandonspark, line 791: would it be better to store this as an `int`?

semgrep_output_v1.atd
```diff
@@ -787,6 +787,13 @@ type api_scans_findings <ocaml attr="deriving show"> = {
...
+  end_line: string;
```
brandonspark, line 792: same here

semgrep_output_v1.atd
```diff
@@ -787,6 +787,13 @@ type api_scans_findings <ocaml attr="deriving show"> = {
    cai_ids: string list;
 }
 
+type finding_hashes <ocaml attr="deriving show"> = {
```
brandonspark, line 790: nit: wording confuses me a bit, is this a hash or plural of hashes? also, this seems to be the information that would go into the hash, rather than the hash itself, if im not mistaken.

README.md
```diff
@@ -9,3 +9,10 @@ rely on this.
 
 This repository is meant to be used as a submodule.
 
+You may need to `install opam` as a pre-requisite for contributing to this repository.
```
brandonspark, line 12: documentation! amazing!

semgrep_output_v1.atd
```diff
@@ -804,6 +811,7 @@ type finding <ocaml attr="deriving show"> = {
   syntactic_id: string;
   (* since semgrep 0.98 *)
   ?match_based_id: string option;
+  ?hashes: finding_hashes option;
```
brandonspark, line 814: im not sure what the use case for this information is, but you may want to add a "since semgrep ..." as above, which will help document how our output interface has evolved

semgrep_output_v1.atd
brandonspark: oooh. or maybe i misread, and this is the entire line itself.

brandonspark (approved):
otherwise, looks good to me! feel free to cc me on the corresponding `semgrep` PR, im interested in what this is being used for

semgrep_output_v1.atd
brandonspark: i now understand this and below as hashes _of the things they are named after_, hence the name `finding_hashes` for the type. it seems i was being silly :)

semgrep_output_v1.atd
brandonspark: i might still recommend a `_hash` prefix after each of these fields, for dummies like me who may get confused :)

dev.opam
aryx: yes, need quite

semgrep_output_v1.atd
```diff
@@ -787,6 +787,13 @@ type api_scans_findings <ocaml attr="deriving show"> = {
    cai_ids: string list;
 }
 
+type finding_hashes <ocaml attr="deriving show"> = {
+  start_line: string;
+  end_line: string;
+  code: string;
```
aryx, line 793: need some comment about what code and pattern are.

dev.opam
chmccreery:
```suggestion
  "atdgen" {>= "2.10.0"}
```

chmccreery: Good catch!

semgrep_output_v1.atd
chmccreery: Good call! Naming is hard

semgrep_output_v1.atd
chmccreery: left my best-guess at when this will get introduced, but will try to remember to adjust it once it's actually released in semgrep

semgrep_output_v1.atd
chmccreery: done!

PR #61, opened 2023-02-27
feat(schema): explicit regex introduction

brandonspark:
This PR is a companion to https://github.com/returntocorp/semgrep/pull/7227, which introduces explicit syntax-level introduction for regex capture group metavariables.


rule_schema_v1.yaml
```diff
@@ -759,6 +765,9 @@ properties:
         match:
           $ref: "#/$defs/new-pattern"
         # EXPERIMENTAL
+        metavars:
```
aryx, line 768: it's at the same level than a match:?

brandonspark: yes, it may occur at the outermost (if the outermost pattern is a `pattern-regex`)

like:
```
rules:
  - id: foo
    languages: [python]
    severity: ERROR
    message: "FOO"
    pattern-regex: "(.*)"
    metavars:
      - 1: $A
```

brandonspark:
Closed in favor of https://linear.app/r2c/issue/PA-2765/analyze-named-capture-groups-to-introduce-regex-metavars

PR #62, opened 2023-03-11
Use ATD's nullable type instead of option

mjambon:
because atdpy doesn't support it correctly and we don't want to use a proper option type in Python anyway.

Using `nullable` instead of `option` in the ATD file will cause `semgrep-core -generate_ast_json` to produce JSON with `null` instead of `"None"`.

This PR adds the generated `.py` file. It requires atdpy version 2.11.0 which is incompatible with the opam dependencies we use for building semgrep-core :-/

In the semgrep repo, which should add a minimal test to check that we can read an AST produced by semgrep-core from a Python program.


aryx:
Ok I'll add the comment in one of my PR on semgrep-interface.

mjambon:
> "Do not use the option type in non-optional fields. Prefer nullable if possible.
For example ?foo: int option; is fine, but not foo: int option; Prefer in that case
foo: int nullable.

I don't think such a comment is the right approach. I mean it's fine but a little misleading because the same remark is true of all ATD files.

Starting with the next release, atdpy will do the right thing (see https://github.com/ahrefs/atd/pull/333). Using `option` will work, it's only that the JSON data will look weirder than if it were using `null`.

PR #63, opened 2023-03-15
do not require `mode: taint` in new style taint rule

brandonspark:
This PR just makes it so `mode: taint` is not required in new-style taint rules.

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -810,9 +810,13 @@ properties:
               pattern-sanitizers: false
               join: false
         - if:
+          # We should only proceed with this old-style taint rule if we do
+          # not see a `taint` top-level key.
+          # Otherwise, it's a new-style taint rule. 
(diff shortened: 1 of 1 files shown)


brandonspark:
It seems I have some of your changes @mjambon, it looks like they weren't pushed to the main repo.

aryx:
Please make a PR with just your change.

aryx (changes requested):
remove the stuff from martin in this PR.

brandonspark:
@aryx is it not the case that even if i did, we wouldn't be able to update our version of `semgrep-interfaces` within `semgrep` because there would be a conflict?

brandonspark:
well, i guess we'll just keep the unmerged changes on the `semgrep` side. i suppose at some point we should push them to `semgrep-interfaces`, though.

aryx (approved):

PR #67, opened 2023-05-03
added Composer ecosystem for ssc

jkylekelly:


aryx:
You need to also call 'make' to update the generated files

aryx (approved):

PR #69, opened 2023-05-04
Undo the parts of commit dd43181 that trigger atdpy bug

mjambon:
atdpy bug: https://github.com/ahrefs/atd/issues/339

I'll let @jkylekelly test if it works better now.

The intention of these changes is to avoid the use of `~xxx: ... list` because of a code-generation bug in atdpy. What doesn't work:
```
~skipped_targets <json name="skipped">: skipped_target list;
```
what is known to work:
```
?skipped_targets <json name="skipped">: skipped_target list option;
```

`~` introduces an optional field that has a default value. In the case of a list, the default value is an empty list unless some other default is specified. The generated Python code that handles this is currently broken.

`?` introduces an optional field that doesn't have a default value. This is what we used to use and it's known to work.


mjambon:
Merging this. Let me know if it still doesn't work.

jkylekelly:
@mjambon getting the same error.  Looks like you've seen this based on slack messages, but for note purposes I've included my replication steps below:

```
git clone git@github.com:returntocorp/semgrep.git
cd semgrep
make dev-setup
cd cli
SEMGREP_SKIP_BIN=true python -m pipenv install --dev
cd /src/semgrep/semgrep_interfaces
make clean
make
cd -
pipenv run pytest tests/e2e/test_dependency_aware_rules.py --snapshot-update
```

PR #71, opened 2023-05-10
Add semgrep_metrics as a common interface

dinosaure:
This PR adds a new `*.atd` file which already exists into the `semgrep` project. The goal is to share `metrics` types between the Python and the OCaml implementation.


semgrep_metrics.atd
```diff
@@ -0,0 +1,138 @@
+(* Semgrep metrics.
+ *
+ * The goal of this module is to specify the schema for the metrics
+ * data sent by the Semgrep CLI to metrics.semgrep.dev.
+ *
+ * translated originally from semgrep/cli/.../metrics.py
+ *
+ * TODO:
+ *  - at some point we should move this file to semgrep-interfaces/
```
aryx, line 9: I guess you can update the TODO

semgrep_metrics.atd
```diff
@@ -0,0 +1,138 @@
...
+(*****************************************************************************)
+
+type payload = {
+    inherit toplevel;
+    environment: environment;
+    performance: performance;
+    extension : extension;
+    errors: errors;
+    value: value;
+    ~parse_rate <python default="field(default_factory=list)">: (string * parse_stat) list <json repr="object">;
```
aryx, line 26: Don't know what @mjambon think, but I'm not a huge fan of those ~xxx fields. I'd rather
have ?xxx or just xxx; IMHO ~xxx add complications without much benefits.

aryx: Or is it so that the python code remains similar to what was before?

aryx (approved):

semgrep_metrics.atd
dinosaure: Yes, if I don't use `~`, I must explicitely define such field when we call the Python constructor. Note that I don't have strong opinion about that and if you think that we should update the Python code to delete on this side the `~` value, I'm fine with that :+1:.

mjambon: `~` means there's a default value. The default value for a list is the empty list, which is broken in the current version of atdpy (2.11). See https://github.com/ahrefs/atd/issues/339

If using `field(default_factory=list)` as a default value works where `[]` doesn't, then let's implement this as the fix in atdpy.

mjambon: It's fine to not use `~`, of course. It's supposed to be more convenient than not using it (when it works).

semgrep_metrics.atd
aryx: Add a TODO maybe that we could remove some of those python default=xxx once atdpy has been fixed

aryx (approved):

PR #72, opened 2023-05-16
Introduce fpath alias for filenames and paths

aryx:


aryx:
This may have consequences on semgrep-app next time you pull the latest semgrep-interfaces @zyannes

PR #73, opened 2023-05-19
Allow extract rules to define paths

emjin:


aryx:
What is that? Can you add more context on this PR?

emjin:
Filed an issue based on my conversations with Lewis: https://github.com/returntocorp/semgrep/issues/7858

PR #74, opened 2023-05-25
validate "id"

jbergler:
Based on current data we only have 8 rules in registry that would fail this validation here and only one customer has rules that fail the validation in their configuration who I will work with to get their config sorted out.


brandonspark (approved):
cool!

PR #79, opened 2023-05-31
Add semgrep_metrics.py to the auto-generated targets

mjambon:
The atdpy bug was fixed in 2.12.0. I'm not sure how the previous version of this file was created. There's a nonempty diff.

Update: the field `maxMemoryBytes` was using `option` but was not optional. We should not use this. I made the field optional.

This updates the version constraints in `dev.opam` and stresses the importance of running `make setup`.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -71,7 +71,7 @@ type performance = {
     ?fileStats : file_stats list option;
     ?ruleStats : rule_stats list option;
     ?profilingTimes : (string * float) list <json repr="object"> option;
-    maxMemoryBytes: int option;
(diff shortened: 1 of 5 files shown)


semgrep_metrics.atd
```diff
@@ -71,7 +71,7 @@ type performance = {
     ?fileStats : file_stats list option;
     ?ruleStats : rule_stats list option;
     ?profilingTimes : (string * float) list <json repr="object"> option;
-    maxMemoryBytes: int option;
+    ?maxMemoryBytes: int option;
   }
```
mjambon, line 75: Please confirm that this is what we want.

PR #81, opened 2023-06-12
Generate accessors for metrics in OCaml

dinosaure:

This PR adds a new ppx to the metrics records to make them easier to handle (and in particular to update). I've tried using `ppx_accessor` to avoid adding dependencies, but it seems that it hasn't kept up with the latest JaneStreet release cycle (Jst's ppx currently have a v0.16.0 version, while `ppx_accessor` remains at v0.15.0 - which is incompatible with semgrep). So I rebuilt an optics library that was useful for other contexts within the OCaml ecosystem anyway: [ppx_lun](https://github.com/roburio/lun).


aryx (approved):
@mjambon is there a way to globally declare the set of deriving we want for all the types defined in a file? so we don't have to repeat ourselves?

aryx:
ah actually I've found it! you can add the "mutable" attribute to a field in ATD!

aryx:
Would be nicer to make it also a global option, so one does not have to repeat himself for every fields ...

aryx (changes requested):
I think it's simpler to not use lun and instead use mutable fields for the metrics.

PR #82, opened 2023-06-13
chore: make cai_ids optonal list

chmccreery:

The cai_ids have not been used for quite some time. I am trying to finish removing them from the app and CLI to simplify the logic in both.

See https://github.com/returntocorp/semgrep-app/pull/8773

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -816,7 +816,7 @@ type api_scans_findings <ocaml attr="deriving show"> = {
    searched_paths: string list;
    (* TODO: rule_id list *)
    rule_ids: string list;
-   cai_ids: string list;
+   cai_ids: string list option;
 }
 
 type finding_hashes <ocaml attr="deriving show"> = {
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -816,7 +816,6 @@ type api_scans_findings <ocaml attr="deriving show"> = {
    searched_paths: string list;
    (* TODO: rule_id list *)
    rule_ids: string list;
-   cai_ids: string list option;
 }
 
 type finding_hashes <ocaml attr="deriving show"> = {
--- a/semgrep_output_v1.jsonschema
+++ b/semgrep_output_v1.jsonschema
@@ -703,8 +703,7 @@
     "api_scans_findings": {
       "type": "object",
       "required": [
-        "findings", "token", "gitlab_token", "searched_paths", "rule_ids",
-        "cai_ids"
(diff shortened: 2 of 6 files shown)


semgrep_output_v1.atd
```diff
@@ -816,7 +816,7 @@ type api_scans_findings <ocaml attr="deriving show"> = {
    searched_paths: string list;
    (* TODO: rule_id list *)
    rule_ids: string list;
-   cai_ids: string list;
+   cai_ids: string list option;
```
aryx, line 819: I think you can just remove the field.
See https://atd.readthedocs.io/en/latest/atdgen.html#smooth-protocol-upgrades
@mjambon what do you think?
The default behavior of ATD as a consumer is that if some JSON data have this cai_ids, and semgrep_output_v1.atd does not have one, it will just be ignored.

aryx: Otherwise when an ATD field is an option, you usually want ?cai_ids: string list option;
meaning this field may even be not present at all in the JSON, and it's ok.
with cai_ids: string list option, that means the JSON will have to contain { ... "cai_ids": null ... }
and that's probably not what we want here.
But again, I think we can just simply remove the field.

chmccreery: Oh, excellent! Yeah, I would much rather remove it if we don't need it, I just wanted to make sure I wouldn't break old versions of semgrep. I'll make that change now!

chmccreery: Ok, I did as you suggested and removed it, but just to be sure: are you sure that old versions of the CLI (e.g. people who are running 1.9.0 or something) won't break because those versions of the CLI expect this field?

aryx (approved):

semgrep_output_v1.atd
aryx: I'll update the CLI tomorrow to use the latest semgrep-interfaces.

aryx: you'll have to do the same for the app backend (and other places maybe too).

chmccreery: Oh, ok, thanks for explaining that. I definitely thought that this was being used to send rule ids from the app to the CLI and not the other way around. I actually don't see it in the `semgrep` repository at all, except in `Unit_reporting.ml`, which is interesting. Maybe it is only being consumed by some analytics server?

aryx: Actually maybe it is the case ... there are many places of pysemgrep i dont know. I just assumed that because it was in semgrep_output.atd it was part of the output of semgrep ... but i ll double check tomorrow when i ll try to update

PR #87, opened 2023-07-07
add metavariable-type to schema

ihji:


ihji:
@brandonspark friendly ping

rule_schema_v1.yaml
```diff
@@ -476,6 +477,27 @@ $defs:
     required:
       - metavariable-pattern
     additionalProperties: false
+  metavariable-type:
+    type: object
+    properties:
+      metavariable-type:
+        type: object
+        title: Match metavariable value with its type
```
brandonspark, line 485: i would write something like "Filter for metavariables with a certain type"

brandonspark (approved):
looks good!

PR #88, opened 2023-07-08
update api_scans_findings to match current usage

jbergler:
This is needed so we can start strongly typing the messages between the CLI and semgrep.dev.

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -811,12 +811,14 @@ type found_dependency <ocaml attr="deriving show"> = {
 type api_scans_findings <ocaml attr="deriving show"> = {
   (* TODO: ?version: version option; *)
    findings: finding list;
+   ignores: finding list;   
    token: string nullable;
-   gitlab_token: string nullable;
    searched_paths: string list;
-   (* TODO: rule_id list *)
+   renamed_paths: string list;
    rule_ids: string list;
 }
+
+
 
 type finding_hashes <ocaml attr="deriving show"> = {
   start_line_hash: string;
--- a/semgrep_output_v1.jsonschema
+++ b/semgrep_output_v1.jsonschema
@@ -703,16 +703,21 @@
     "api_scans_findings": {
       "type": "object",
       "required": [
-        "findings", "token", "gitlab_token", "searched_paths", "rule_ids"
+        "findings", "ignores", "token", "searched_paths", "renamed_paths",
+        "rule_ids"
       ],
       "properties": {
         "findings": {
           "type": "array",
           "items": { "$ref": "#/definitions/finding" }
         },
+        "ignores": {
+          "type": "array",
+          "items": { "$ref": "#/definitions/finding" }
+        },
         "token": { "type": [ "string", "null" ] },
-        "gitlab_token": { "type": [ "string", "null" ] },
         "searched_paths": { "type": "array", "items": { "type": "string" } },
+        "renamed_paths": { "type": "array", "items": { "type": "string" } },
         "rule_ids": { "type": "array", "items": { "type": "string" } }
       }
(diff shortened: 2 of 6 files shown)


semgrep_output_v1.atd
```diff
@@ -811,13 +811,15 @@ type found_dependency <ocaml attr="deriving show"> = {
 type api_scans_findings <ocaml attr="deriving show"> = {
   (* TODO: ?version: version option; *)
    findings: finding list;
+   ignores: finding list;   
    token: string nullable;
-   gitlab_token: string nullable;
    searched_paths: string list;
-   (* TODO: rule_id list *)
+   renamed_paths: string list;
```
jbergler, line 817: Do we need to double check this field? It looks like it was moved around a bit but I think it's a no-op.
```
    ⋮ 301│        api_scans_findings = out.ApiScansFindings(
 282⋮ 302│            # send a backup token in case the app is not available
 283⋮    │            "token": os.getenv("GITHUB_TOKEN"),
 284⋮    │            "gitlab_token": os.getenv("GITLAB_TOKEN"),
 285⋮    │            "findings": findings,
 286⋮    │            "searched_paths": [str(t) for t in targets],
 287⋮    │            "renamed_paths": [str(rt) for rt in renamed_targets],
 288⋮    │            "rule_ids": rule_ids,
 289⋮    │            "cai_ids": cai_ids,
    ⋮ 303│            token=token,
    ⋮ 304│            findings=findings,
    ⋮ 305│            searched_paths=[str(t) for t in sorted(targets)],
    ⋮ 306│            rule_ids=rule_ids,
    ⋮ 307│            gitlab_token=None,
    ⋮ 308│        )
    ⋮ 309│        # TODO: add those fields in semgrep_output_v1.atd spec
    ⋮ 310│        findings_and_ignores = {
    ⋮ 311│            **api_scans_findings.to_json(),
    ⋮ 312│            "renamed_paths": [str(rt) for rt in sorted(renamed_targets)],
 290⋮ 313│            "ignores": ignores,
 291⋮ 314│        }
```

aryx:
I have no idea. Maybe @chmccreery knows about this renamed_path field.

jbergler:
> I have no idea. Maybe @chmccreery knows about this renamed_path field.

All good, the question was for @mmcqd

semgrep_output_v1.atd
mmcqd: This field has been included in the json sent to `findings_and_ignores` since at least 0.123.0: https://github.com/returntocorp/semgrep/blob/084ff5f16866ab66337655dbc06672e05d3089dc/cli/src/semgrep/app/scans.py#L281

It just hasn't been included in the ATD type

jbergler: Discussed in slack, and this is fine

mmcqd: jonas got this diff backwards. Here's the code creating this field in 0.123.0: https://github.com/returntocorp/semgrep/blob/084ff5f16866ab66337655dbc06672e05d3089dc/cli/src/semgrep/app/scans.py#L281

It's hasn't changed type/existence since then, so this change should indeed be a no-op, but will enable us to do more type-checking

mmcqd: cc @aryx, should be good to merge if you approve

jbergler:
Trying to update the CLI to use these changes produces the following build error 
```
File "src/osemgrep/cli_ci/Ci_subcommand.ml", line 261, characters 8-20:
261 |         gitlab_token = None;
              ^^^^^^^^^^^^
Error: Unbound record field gitlab_token
make[2]: *** [minimal-build] Error 1
make[1]: *** [core] Error 2
make: *** [build] Error 2
```

PR #89, opened 2023-07-12
enable the migration of severity from int to string

jbergler:
This should be an string enum rather than some int representation that approximates it.
To migrate we need the typing to allow for both string and int, which we can do by pushing the typing up a layer until the CLI versions using int are deprecated.

Closes APP-4267

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -105,7 +105,7 @@ type rule_id
 
 type rule_id_and_engine_kind
   <ocaml attr="deriving show">
-  <python decorator="dataclass(frozen=True)"> = (rule_id * engine_kind) 
+  <python decorator="dataclass(frozen=True)"> = (rule_id * engine_kind)
 
 (*****************************************************************************)
 (* Core Match result *)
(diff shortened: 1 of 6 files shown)


aryx:
I d rather have an enum. A string is ugly and unsafe.

jbergler:
> I d rather have an enum. A string is ugly and unsafe.

Agreed, but we're hot swapping this out so it's going to be a multi-step dance.
My first goal is to get the cli sending the string values that match the rule schema, and then when we're 10+ versions of the CLI on and we can fully deprecate the int values we can update the atd to match.

mjambon (approved):
This should work. I don't have a better suggestion.

semgrep_output_v1.atd
```diff
@@ -836,8 +836,7 @@ type finding <ocaml attr="deriving show"> = {
   end_line: int;
   end_column: int;
   message: string;
-  (* TODO: use variant *)
-  severity: int;
+  severity: abstract; (* int|string until minimum version exceeds 1.32.0, then string *)
```
mjambon, line 839: What are valid values for the int and for the string, and how to convert one from the other? Is it documented somewhere else?

PR #90, opened 2023-07-14
add metavariable-type syntax to rule 2.0

ihji:


nmote (approved):
Looks good to me!

PR #91, opened 2023-07-14
Add .h for c & c++

ievans:
Not sure if I need to run something after updating this

### DIFF
--- a/lang.json
+++ b/lang.json
@@ -48,7 +48,8 @@
     ],
     "maturity": "alpha",
     "exts": [
-      ".c"
+      ".c",
+      ".h"
     ],
     "example_ext": null,
     "excluded_exts": [],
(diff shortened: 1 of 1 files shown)


emjin (approved):
Seems reasonable but I'd like to see the effect on our parse rate once this change is pulled in. It would be great if we could also have a test to make sure we parse .h files using c!

lang.json
```diff
@@ -104,7 +105,8 @@
     "maturity": "alpha",
     "exts": [
       ".cc",
-      ".cpp"
+      ".cpp",
+      ".h"
```
jbergler, line 109: Shouldn't this be `.hh` ?

aryx: many C++ projects also use .h for their header.
But yes that means we will parse .h files twice, once for C, and once for C++.
But yes we could also add .hh

aryx:
@emjin lang.json is now autogenerated from generate.py, so we should edit generate.py instead of lang.json!
We can't have comments in json, but maybe @mjambon we can add a fake first entry with a big "DO_NOT_EDIT_THIS_FILE"

PR #92, opened 2023-07-14
langs: Add .cjs and .mjs for javascript

underyx:


emjin (approved):
Same comment as for Isaac; it seems reasonable but I'd like to see the effect on our parse rate once this change is pulled in. And it would be great if we could also have a test to make sure we parse .cjs files using javacript!

PR #93, opened 2023-07-15
feat(promql): add language

MichaHoffmann:


aryx:
I think you need to rebase on develop because you should not remove the unrelated .h extension that was added recently for C and C++.

lang.json
```diff
@@ -48,8 +48,7 @@
     ],
     "maturity": "alpha",
     "exts": [
-      ".c",
```
aryx, line 51: This should not be part of the PR.
You should probably rebase on main (or develop, whatever the master branch is).

aryx (changes requested):

lang.json
MichaHoffmann: not sure what can be done; the commit that added those filetypes added them to this autogenerated file and i autogenerated it away. Ill at a commit to this PR to add them properly to `generate.py`

aryx:
This also fix the error in https://github.com/returntocorp/semgrep-interfaces/pull/91

aryx (approved):

PR #96, opened 2023-07-20
Add type for `/complete` data

mmcqd:
All the optional fields are new since `0.123.0`

0.123.0 data: https://github.com/returntocorp/semgrep/blob/084ff5f16866ab66337655dbc06672e05d3089dc/cli/src/semgrep/app/scans.py#L301

Current data: https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/app/scans.py#L320

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -801,6 +801,39 @@ type found_dependency <ocaml attr="deriving show"> = {
   ?line_number: int option;
 }
 
+
+
+type sca_parser_name = [
+  | Gemfile_lock <json name="gemfile_lock">
+  | Go_mod <json name="go_mod">
+  | Go_sum <json name="go_sum">
+  | Gradle_lockfile <json name="gradle_lockfile">
+  | Gradle_build <json name="gradle_build">
+  | Jsondoc <json name="jsondoc">
+  | Pipfile <json name="pipfile">
+  | Pnpm_lock <json name="pnpm_lock">
+  | Poetry_lock <json name="poetry_lock">
+  | Pyproject_toml <json name="pyproject_toml">
+  | Requirements <json name="requirements">
+  | Yarn_1 <json name="yarn_1">
+  | Yarn_2 <json name="yarn_2">
+  | Pomtree <json name="pomtree">
+  | Cargo_parser <json name="cargo">
+  | Composer_lock <json name="composer_lock">
+]
+
+
+
+type dependency_parser_error = {
+  path: string;
+  parser: sca_parser_name;
+  reason: string;
+  ?line: int option; 
+  ?col: int option; 
+  ?text: string option;
+}
+
+
 (*****************************************************************************)
 (* Semgrep CLI findings output for the App *)
 (*****************************************************************************)
@@ -808,6 +841,7 @@ type found_dependency <ocaml attr="deriving show"> = {
  * there is also api_scans_meta, api_scans_ignores, and api_scans_complete
  * EXPERIMENTAL
 *)
+
 type ci_scan_results <ocaml attr="deriving show"> = {
   (* TODO: ?version: version option; *)
    findings: finding list;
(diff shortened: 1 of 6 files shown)


semgrep_output_v1.atd
```diff
@@ -820,6 +854,47 @@ type ci_scan_results <ocaml attr="deriving show"> = {
 
 
 
+type parsing_stats <ocaml attr="deriving show"> = {
```
jbergler, line 857: I don't know a lot about atd, but I think it supports includes right? Would it make sense to split the product stuff out into domain specific files and include things?

semgrep_output_v1.atd
```diff
@@ -820,6 +854,47 @@ type ci_scan_results <ocaml attr="deriving show"> = {
...
+  (* This is EngineType from python, which is different from engine_kind used in this file *)
+  ?engine_requested: string option;
+}
+
+type ci_scan_dependencies <ocaml attr="deriving show"> =  (string * found_dependency list) list
+    <json repr="object">
+    <python repr="dict">
+    <ts repr="map">
+
+type ci_scan_complete <ocaml attr="deriving show"> = {
```
jbergler, line 889: should we try to be more specific about req vs resp, eg `ci_scan_complete_response` or something?

jbergler (approved):

semgrep_output_v1.atd
aryx: The current ATD does not support include (it was originally a design decision of @mjambon. Since then martin
started to add support for includes/modules, but it's not merged yet.
But yes would be good to split semgrep_output_v1.atd in multiple files at some point.

aryx (approved):

semgrep_output_v1.atd
mmcqd: It supports include between records within the same file. As far as I can tell support for imports across files is only available when generating ocaml, it doesn't work in python: https://atd.readthedocs.io/en/latest/atdgen.html#modularity-referring-to-type-definitions-from-another-atd-file

semgrep_output_v1.atd
mmcqd: Yeah that seems reasonable

semgrep_output_v1.atd
```diff
@@ -801,13 +801,47 @@ type found_dependency <ocaml attr="deriving show"> = {
...
+  | Composer_lock <json name="composer_lock">
+]
+
+
+
+type dependency_parser_error = {
+  path: string;
+  parser: sca_parser_name;
+  reason: string;
+  ?line: int option;
```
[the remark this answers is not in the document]
mmcqd: Good idea. I just copied the python definition we had but that would be better

semgrep_output_v1.atd
aryx: I was referring to: https://github.com/ahrefs/atd/pull/297

semgrep_output_v1.atd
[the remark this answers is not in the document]
mmcqd: Turns out we can't do that, because the app needs to be able to deserialize data that came from the original python class that this is replacing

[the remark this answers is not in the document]
aryx: Then please add a comment about it in the .atd file (why we can't simply reuse 'position';
also I don't fully understand your explanation so maybe a deeper explanation would be useful).

semgrep_output_v1.atd
```diff
@@ -801,13 +801,48 @@ type found_dependency <ocaml attr="deriving show"> = {
   ?line_number: int option;
 }
 
+
+(* json names are to maintain backwards compatibility with the python enum it is replacing *)
```
aryx, line 805: great comment!

aryx:
I see the PR reference /complete but the .atd does not talk about it.
Maybe we could add comments in the .atd for the different "entry" types and specify
to which endpoint they correspond to?

semgrep_output_v1.atd
```diff
@@ -801,13 +801,48 @@ type found_dependency <ocaml attr="deriving show"> = {
...
+  | Composer_lock <json name="composer_lock">
+]
+
+
+
+type dependency_parser_error <ocaml attr="deriving show"> = {
+  path: string;
+  parser: sca_parser_name;
+  reason: string;
+  (* Not using `position` because this type must be backwards compatible with the python class it is replacing *)
```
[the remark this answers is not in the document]
mmcqd: We can refactor the python code, and I agree we should. The main difficulty is with supporting old versions of the CLI. We have no choice but to support this type for a while. 

Would be really cool if ATD had some kind of "compatibility sum type", where I could say "it's either this or this, and ATD should insert the tag depending on the shape of the data". Like

```
type sca_parser_error <compat-sum> = [
  | OldType of old_type
  | NewType of new_type
]
```

And then the actual data is untagged, since old versions would just be sending `old_type` by itself

mmcqd:
> I see the PR reference /complete but the .atd does not talk about it. Maybe we could add comments in the .atd for the different "entry" types and specify to which endpoint they correspond to?

Done

aryx:
You should talk with @mjambon when he is back from vacation about your ideas for improving ATD.

PR #97, opened 2023-07-20
chore: Add contribution model

zyannes:

For context, [this PR](https://github.com/returntocorp/semgrep/pull/7892) was created a while back to get the "contributions" in a CI scan and upload with the findings/ignores, but it was implemented in python.

I'm translating that logic to semgrep-core so it will be available to both pysemgrep and osemgrep.

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -464,7 +464,6 @@ type core_match_results
   (* since semgrep 0.109? *)
   ?explanations: matching_explanation list option;
 
-
   stats: core_stats;
   (* LATER: rename timing *)
   ?time: core_timing option;
@@ -713,6 +712,17 @@ type cli_timing <ocaml attr="deriving show"> = {
     ?max_memory_bytes : int option;
   }
 
+type contribution <ocaml attr="deriving show"> = {
+    commit_hash: string;
+    commit_timestamp: string;
+    commit_author_name: string;
+    commit_author_email: string;
+}
+
+type contributions <ocaml attr="deriving show"> = {
+    contributions: contribution list;
+}
+
 (* LATER: should just use rule_id *)
 type rule_id_dict <ocaml attr="deriving show"> = {
   id: rule_id;
@@ -816,6 +826,7 @@ type ci_scan_results <ocaml attr="deriving show"> = {
    searched_paths: string list;
    renamed_paths: string list;
    rule_ids: string list;
+   ?contributions: contributions option;
 }
 
 
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -712,11 +712,18 @@ type cli_timing <ocaml attr="deriving show"> = {
     ?max_memory_bytes : int option;
   }
 
+(* 
+ * https://semgrep.dev/docs/usage-limits
+ *)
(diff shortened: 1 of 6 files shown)


semgrep_output_v1.atd
```diff
@@ -713,6 +712,17 @@ type cli_timing <ocaml attr="deriving show"> = {
     ?max_memory_bytes : int option;
   }
 
+type contribution <ocaml attr="deriving show"> = {
```
aryx, line 715: maybe some link to some documentation about what a "contribution" means, or an example.

semgrep_output_v1.atd
```diff
@@ -816,6 +826,7 @@ type ci_scan_results <ocaml attr="deriving show"> = {
    searched_paths: string list;
    renamed_paths: string list;
    rule_ids: string list;
+   ?contributions: contributions option;
```
aryx, line 829: We usually try to say since which version those additional fields have been supported, like
 (* since semgrep 1.22.0 *)
for example.

semgrep_output_v1.atd
```diff
@@ -713,6 +712,17 @@ type cli_timing <ocaml attr="deriving show"> = {
     ?max_memory_bytes : int option;
   }
 
+type contribution <ocaml attr="deriving show"> = {
+    commit_hash: string;
+    commit_timestamp: string;
+    commit_author_name: string;
```
aryx, line 718: maybe this could have been split in a separate contributor type, like it was done in the Python code of shapski?

aryx (approved):

aryx:
Also thanks to those ATD definitions, I was hoping in addition to be supported both in osemgrep and pysemgrep, it would
also be supported in the backend in a typed way?

semgrep_output_v1.atd
zyannes: I optimistically set it for 1.34.0. (I figure I can go back and update it if the semgrep-core code doesn't make it into the next release)

semgrep_output_v1.atd
zyannes: Added a link to the public usage-limit docs. I suspect we'll update those docs at some point soon with more details about how the contributor/contribution count works.

semgrep_output_v1.atd
zyannes: Good call! Added in latest commit

semgrep_output_v1.atd
aryx: Oh wow, I completely misinterpreted what it was for at the beginning. I thought it was just for some internal stats about who was using semgrep (well it is, but the objective is now clearer).

aryx: How usage limit was enforced before if we were not sending this info? Or we were sending it but it's just it was not specified in the ATD file?

zyannes: Before this, we would send the commit info (commit hash, author name, author email) as scan metadata when the CLI first created the scan. However, this would only account for a single commit, and it could be overwritten by environment variables (those had a higher priority).

The new approach is to also use `git log` as a source of truth. I say "also" because the CLI will still have the option to configure the latest commit hash/author/email with environment variables.

PR #103, opened 2023-07-28
feat(...): extend rule schema for internal post-processor

yosefAlsuhaibani:

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -957,6 +957,7 @@ properties:
               patterns: false
               pattern: false
               pattern-either: false
+              pattern-regex: false
               pattern-sinks: false
               pattern-sources: false
               pattern-propagators: false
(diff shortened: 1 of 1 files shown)


rule_schema_v1.yaml
```diff
@@ -953,7 +957,6 @@ properties:
               patterns: false
               pattern: false
               pattern-either: false
-              pattern-regex: false
```
aryx, line 956: Is this related to the post-processor change?

aryx (approved):

rule_schema_v1.yaml
yosefAlsuhaibani: not quite; restored!

aryx:
BTW it's usually good after you make a PR to have yourself a quick look at the diff, to fix simple things like that.
You should be your first reviewer.

PR #104, opened 2023-08-01
generate protobuf definition for semgrep_output_v1

tpetr:


scripts/jsonschema2protobuf.py
```diff
@@ -0,0 +1,214 @@
+#!/usr/bin/env python3
```
aryx, line 1: Where does this script comes from? What is the copyright? Author?
A toplevel comment explaining those things, limitations, etc. would be useful.

aryx (changes requested):

scripts/jsonschema2protobuf.py
tpetr: Added

ajbt200128 (approved):

aryx (approved):

PR #106, opened 2023-08-02
Hack week refinments to semgrep_studio_v1.proto NO NEED TO MERGE

aryx:
test plan:
nope


Makefile
```diff
@@ -15,7 +15,7 @@ FILES= \
   semgrep_output_$(VER).py \
   semgrep_output_$(VER).ts \
   semgrep_output_$(VER).jsonschema \
-  semgrep_output_$(VER).proto \
+  protos/semgrep_output_$(VER).proto \
```
aryx, line 18: why do we need its own folder? Can you add a README or notes.txt in the protos/ folder why we need that?

protos/semgrep_studio_v1.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+package semgrep_studio_v1;
+
+option go_package = "semgrep_studio.v1";
```
aryx, line 5: why do we need this? Can you add a comment about it?
Do we use Go somewhere at Semgrep Inc. ?

Makefile
tpetr: Whoops, thanks for the reminder. I think originally moved this into `protos/` so that it'd play nice with `semgrep-app` (it expects protos to live inside a protos folder), but I'm not sure if this is necessary anymore. Let me refresh my memory and tweak or comment accordingly. (cc @ajbt200128)

protos/semgrep_studio_v1.proto
tpetr: Yes, `semgrep-app` is configured to generate golang types, and `go_package` is a requirement for that to work. It might be simpler to just exempt this file, though

Makefile
ajbt200128: App worked w/out the extra protos dir I believe, so we should be fine on that end

protos/semgrep_studio_v1.proto
aryx: Do we use Go in semgrep-app??

aryx:
old stuff, let's close it.

PR #108, opened 2023-08-10
chore(finding): add optional product field to finding

pangngan:


emjin (approved):

semgrep_output_v1.atd
```diff
@@ -956,4 +956,5 @@ type finding <ocaml attr="deriving show"> = {
   ?fixed_lines: string list option;
   ?sca_info: sca_info option;
   ?dataflow_trace: cli_match_dataflow_trace option;
+  ?product: string option;
```
aryx, line 959: it's useful to also add in comment since which version of semgrep the field was introduced.

PR #113, opened 2023-08-30
Add support for min-version/max-version and incompatible rule reporting

mjambon:
* This adds a new kind of error called "IncompatibleRule".
* This also adds a new field `skipped_rules` to the semgrep response but it remains unpopulated in the implementation that's about to be merged.
* There's also the addition of a new error level called "info".
* The position type used by semgrep-core and semgrep is now the same.

The implementation: https://github.com/returntocorp/semgrep/pull/8496


semgrep_output_v1.atd
```diff
@@ -228,14 +235,30 @@ type core_error_kind
...
   (* since semgrep 0.97 *)
   | PartialParsing of location list (* list of skipped tokens *)
+  (* since semgrep 1.37.0 *)
+  | IncompatibleRule of incompatible_rule
 ] <ocaml repr="classic">
 
+type incompatible_rule
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  rule_id: string;
```
aryx, line 245: there is a rule_id type defined in this file you can use instead of string.

semgrep_output_v1.atd
```diff
@@ -64,14 +64,21 @@ type version <ocaml attr="deriving show"> = string (* e.g., "1.1.0" *)
 (*****************************************************************************)
 (* Location *)
 (*****************************************************************************)
-(* note that there is no filename here like in 'location' below *)
+(* Note that there is no filename here like in 'location' below.
+
+   Upgrade note: "position_bis" was merged with "position".
```
aryx, line 69: nice!

semgrep_output_v1.atd
```diff
@@ -228,14 +235,30 @@ type core_error_kind
...
+  (* since semgrep 1.37.0 *)
+  | IncompatibleRule of incompatible_rule
 ] <ocaml repr="classic">
 
+type incompatible_rule
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = {
+  rule_id: string;
+  this_version: string;
+  ?min_version: string option;
```
aryx, line 247: There is also a version type defined in this file you can use instead of string.
It's an alias for a string, but we could change that in the futur with more structure (e.g., major, minor, patch)

semgrep_output_v1.atd
```diff
@@ -686,6 +703,9 @@ type cli_output_extra <ocaml attr="deriving show"> = {
     (* EXPERIMENTAL: since: 1.11.0 *)
     ?rules_by_engine: rule_id_and_engine_kind list option;
     ?engine_requested: engine_kind option;
+
+    (* EXPERIMENTAL: since: 1.37.0 *)
+    skipped_rules: skipped_rule list;
```
aryx, line 708: why not a ? before so new consumers can still read stuff from old producers.

aryx (approved):

semgrep_output_v1.atd
mjambon: This is a mistake. Not sure what my reasoning was. I'll leave the field required in the semgrep-core output, though.

mjambon: Ah yes, I wanted to avoid a discrepancy between pysemgrep and osemgrep output. The more convenient way to define fields that default to the empty list is `~` rather than `?` (defaults to empty list vs. defaults to None) but it produces different JSON with atdpy and atdgen. The latter tries to be smart and omits the field if its value is equal to the default, causing `"skipped_rules": []` in the JSON output from Python vs. <no field> in the JSON output from OCaml.

I'll try `atdgen -j-defaults` to match the atdpy behavior. If this doesn't work, `?` should work.

mjambon: I'm happy to report `-j-defaults` works. This illustrates the complexity of atdgen. atdpy is much simpler and saner.

semgrep_output_v1.atd
```diff
@@ -686,6 +703,9 @@ type cli_output_extra <ocaml attr="deriving show"> = {
     (* EXPERIMENTAL: since: 1.11.0 *)
     ?rules_by_engine: rule_id_and_engine_kind list option;
     ?engine_requested: engine_kind option;
+
+    (* EXPERIMENTAL: since: 1.37.0 *)
+    ~skipped_rules: skipped_rule list;
```
aryx, line 708: I'm still confused about ~. I understand ?, but each time I see ~ I forget what it does.

aryx (approved):

semgrep_output_v1.atd
mjambon: `~` means a default value is used if the field is missing. See https://atd.readthedocs.io/en/latest/atdpy.html#defining-default-field-values

aryx: so when you output the JSON, if skipped_rules is the empty list, will it output a skipped_rules: [] in the JSON?
and when you read the JSON, if there is no skipped_rules in the JSON, it will generate a record with an empty list right? (avoiding some annoying match option)

aryx: I think it would be nice in the doc to give the 2 scenarios, when producing JSON, and when reading JSON, because this has implication on the smooth upgrade process.

mjambon:
ok, the tests pass and I'm done with the fixes in the ATD file.

PR #117, opened 2023-09-04
Update C++ source and header extensions based on tokei

neuroo:
Several extensions were missing for C++. Used [tokei's source](https://github.com/XAMPPRocky/tokei/blob/master/languages.json) and updated the C++ extensions to represent both headers and source files.


generate.py
```diff
@@ -246,7 +246,7 @@ def to_json(self):
         id_="cpp",
         name="C++",
         keys=["cpp", "c++"],
-        exts=[".cc", ".cpp", ".h"],
+        exts=[".cc", ".cpp", ".cxx", ".c++", ".pcc", ".tpp", ".h", ".hh", ".hpp", ".hxx", ".inl", ".ipp"],
```
aryx, line 249: maybe you could add a comment here saying this list was derived from tokei source with a URL link

aryx (approved):

PR #121, opened 2023-09-06
remove cli_ prefix for match_dataflow_trace stuff

aryx:
Now that the cli_xx and core_xx have been merged in one
single type, there is no more need for this cli_ prefix

test plan:
see related PR in semgrep


Makefile
```diff
@@ -82,3 +82,17 @@ setup:
...
+
+pr:
+	git push origin `git rev-parse --abbrev-ref HEAD`
+	hub pull-request -b main -r returntocorp/pa
+
+push:
+	git push origin `git rev-parse --abbrev-ref HEAD`
+
+merge:
+	A=`git rev-parse --abbrev-ref HEAD` && git checkout main && git pull && git branch -D $$A
```
nmote, line 98: Not a huge fan of these, but whatever

nmote (approved):

PR #122, opened 2023-09-07
simplify profiling, merge core_timing with cli_timing

aryx:
Also use better names

test plan:
see related PR in semgrep


semgrep_output_v1.atd
```diff
@@ -554,52 +554,62 @@ type rule_times
...
+type profile <ocaml attr="deriving show"> = {
+    (* List of rules, including the one read but not run on any target.
+     * This list is actually more an array which allows other
+     * fields to reference rule by number instead of rule_id
+     * (e.g., match_times further below) saving space in the JSON.
+     *
+     * Upgrade note: this used to be defined as a rule_id_dict where
+     * each rule_id was inside a {id: rule_id} record (maybe so
+     * we could have given more profiling info about each rule, but
+     * this was never used, so better to make it simple).
```
emjin, line 577: Ah yeah that was because we used to record a parse time for each rule; this is better now. Feel free to delete this note

semgrep_output_v1.atd
```diff
@@ -554,52 +554,62 @@ type rule_times
...
-*)
-
-(* similar to the [target_time] and [rule_times] type above *)
 type cli_target_times <ocaml attr="deriving show"> = {
     path: fpath;
     num_bytes: int;
-    (* each elt in the list refers to a rule in cli_timing.rules *)
+    (* each elt in the list refers to a rule in profile.rules *)
     match_times: float list;
+    (* TODO: why a list? *)
```
emjin, line 610: This doesn't need to be a list anymore. When we were first diagnosing performance, I recorded every time the file was read (including the later times that were just reloading the parsed file) to make sure reading the file wasn't taking a significant amount of time. Now that we know it isn't, we don't need to record this anymore

semgrep_output_v1.atd
```diff
@@ -554,52 +554,62 @@ type rule_times
...
     parse_times: float list;
+    (* run time for all rules on target? *)
```
emjin, line 612: Yes

emjin (commented):
Thanks for the changes! I commented with context where I remember it

PR #123, opened 2023-09-08
cleanup semgrep_output_v1.atd and semgrep_metrics.atd

aryx:
add a few comments too

test plan:
see related PR in semgrep


kopecs:
Looks like some of these deriving shows were needed for pro. :(

PR #128, opened 2023-09-12
add Dotfile for skip_reason

aryx:
test plan:
see related PR in semgrep


semgrep_output_v1.atd
```diff
@@ -449,7 +449,9 @@ type skip_reason <ocaml attr="deriving show"> = [
   | Irrelevant_rule <json name="irrelevant_rule">
   | Too_many_matches <json name="too_many_matches">
   (* New in osemgrep *)
-  | Gitignore_patterns_match <json name="gitignore_patterns_match">
+  | Gitignore_patterns_match
+  (* since 1.40.0 (dotfiles were always ignored, but not shown in the skip report *)
+  | Dotfile
 ] <ocaml repr="classic">
```
mjambon, line 455: Why not stick with lowercase names in JSON? I don't like the inconsistency.

aryx: The previous entries were added to be backward compatible with how things were encoded before, so we needed those ugly redundant <json name="blabla">. For new entry we don't have this constraint so I prefer to keep the code simple so I don't have to bother to add those annoying <json name="blabla">

PR #129, opened 2023-09-12
allow additional properties

brendongo:

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -137,8 +137,8 @@ $defs:
                     title: Integer base to parse metavariable contents as
                     type: integer
                   strip:
-                    title: Whether to strip quotes from compared metavariables 
-                    type: boolean 
+                    title: Whether to strip quotes from compared metavariables
+                    type: boolean
               - required: [ metavariable ]
                 $ref: "#/$defs/new-pattern"
                 properties:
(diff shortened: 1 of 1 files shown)


brendongo:
Fixes https://github.com/returntocorp/semgrep/issues/8648

mjambon (approved):
looks good to me. Let's delay the merge until we have a chance to check the quality of error messages created by the OCaml implementation when an unknown field is encountered. See https://github.com/returntocorp/semgrep/issues/8648#issuecomment-1716665999

aryx:
I think we can merge that; the OCaml implementation does not rely on the rule schema anyway, so we can independently improve things.
@brendongo can you add a big comment though before this important setting, like "# THIS IS IMPORTANT! we allow additional fields because X and Y. That does not mean you can write whatever rule you want, but for backward and forward compatibility reason, we don't want to crash if semgrep runs on rule using deprecated fields, or old semgrep runs on rules using new fields; We can just skip those rules in the engine without crashing at validation time" (or something like that).

mjambon:
@aryx we can continue rejecting unknown fields in semgrep-core/osemgrep. We only need to be tolerant in pysemgrep because it doesn't check for `min-version`. If a new rule uses a newly-introduced field, it should set `min-version` appropriately.

PR #130, opened 2023-09-13
small cleanup, use more rule_id and fpath

aryx:
test plan:
see related PR in semgrep


semgrep_output_v1.atd
```diff
@@ -886,57 +912,68 @@ type ci_scan_complete_stats = {
...
-  (* hash of the rule pattern with metavariables substituted in *)
-  pattern_hash: string;
-}
+(* ----------------------------- *)
+(* Findings *)
+(* ----------------------------- *)
 
-(* TODO: rewrite rule_matches.to_app_finding_format() *)
+(* TODO: rewrite rule_matches.to_app_finding_format()
+ * TODO: what's the endpoint? Where this is sent?
```
zyannes, line 932: This is sent by the CLI to the `/results` endpoint before calling `/complete`. We can have more rigorous contract testing on the app-side to check for backwards compatibility here after a version upgrade.

zyannes (approved):

PR #131, opened 2023-09-13
Define Product shared type and start of the meta type

aryx:
This is the start of typechecking this meta thing
that is used in semgrep ci.

test plan:
see related PR in semgrep


semgrep_output_v1.proto
```diff
@@ -289,6 +289,35 @@ message DependencyParserError {
   string text = 3574323;
 }
 
+message Finding {
```
zyannes, line 292: Since this breaks the backwards compatibility of the `Finding` proto, what do you think about bumping this to `semgrep_output_v2`?

zyannes (approved):

aryx:
why would this break the proto?
I didn't change anything, just moved around things mostly. the JSON for the findings should be exactly the same.

PR #133, opened 2023-09-18
Make offset optional again

mmcqd:
Undoes this commit: https://github.com/returntocorp/semgrep-interfaces/commits/main?after=46309780f7ce3534f2fb89d71666ea207e6eca31+34&branch=main&qualified_name=refs%2Fheads%2Fmain

This field is included in data that is sent to semgrep-app, specifically in the `cli_error` type. Making required causes the app to fail when reading data from old version of semgrep.


jbergler (approved):

semgrep_output_v1.atd
```diff
@@ -80,7 +80,7 @@ type position
   (* Byte position from the beginning of the file, starts at 0.
    * OCaml code sets it correctly. Python code sets it to a dummy value (-1).
    *)
-  offset: int;
+  ~offset: int;
```
aryx, line 83: Could be good to put a deprecated notice like (* Semgrep CLI version <= 1.xx.0 was not always producing an offset, hence the ~, but once we don't support that old version in the App, we can remove the ~ *)

PR #134, opened 2023-09-20
Merge cli_target_times with target_time

aryx:
One step closer getting semgrep-core outputting directly
a cli_output instead of core_output
(so pysemgrep becomes more of an identity function)

test plan:
see related PR in semgrep


aryx:
yes, https://github.com/returntocorp/semgrep/pull/8778

aryx:
and also https://github.com/returntocorp/semgrep/pull/8777

PR #136, opened 2023-09-24
chore: Add runner protos

zyannes:

### DIFF
--- a/semgrep_runner_v1.proto
+++ b/semgrep_runner_v1.proto
@@ -16,9 +16,10 @@ option go_package = "semgrep_runner.v1";
 import "protos/semgrep_output_v1.proto";
 
 enum RuleFormat {
-  YAML = 0;
-  JSON = 1;
-  JSONNET = 2;
+    RULE_FORMAT_UNSPECIFIED = 0;
+    RULE_FORMAT_YAML = 1;
+    RULE_FORMAT_JSON = 2;
+    RULE_FORMAT_JSONNET = 3;
 }
 
 message FilePath {
--- a/semgrep_runner_v1.proto
+++ b/semgrep_runner_v1.proto
@@ -1,0 +1,38 @@
+syntax = "proto3";
+
+package semgrep_runner_v1;
+
+option go_package = "semgrep_runner.v1";
+
+import "protos/semgrep_output_v1.proto";
+
+enum RuleFormat {
+  YAML = 0;
+  JSON = 1;
+  JSONNET = 2;
+}
+
+message FileBlob {
+    string file_path = 1;
+    string contents = 2;
+}
+
+message ScanRequest {
+    // the rule raw content
+    string rule = 1;
+    // how to parse the rule content
+    RuleFormat rule_format = 2;
+    // list of file blobs to receive via SubmitFileBlob service
+    repeated FileBlob files = 3;
+}
+
+message ScanResult {
+    // same as one of the repository_names in the request
+    string request_id = 1;
+    semgrep_output_v1.CliMatch cli_match = 2;
+}
+
+service ScanService {
+    // we provide a streaming interface here! to return results as soon as they arrive!
+    rpc Scan(ScanRequest) returns (stream ScanResult) {};
+}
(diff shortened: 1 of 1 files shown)


semgrep_runner_v1.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
aryx, line 1: Would be nice to have a toplevel comment explaining briefly what is the purpose of this file (you can have a look at semgrep_output_v1.atd for an example, even though it does not have to be that long :) )

aryx: Maybe a small comment explaining also where is the actual server (I imagine, semgrep-pro/server/...) and the actual client (I imagine somewhere in semgrep-app).

semgrep_runner_v1.proto
```diff
@@ -0,0 +1,38 @@
...
+
+package semgrep_runner_v1;
+
+option go_package = "semgrep_runner.v1";
```
aryx, line 5: We use Go? There will be some Go client using the semgrep runner?

semgrep_runner_v1.proto
```diff
@@ -0,0 +1,38 @@
...
+import "protos/semgrep_output_v1.proto";
+
+enum RuleFormat {
+  YAML = 0;
+  JSON = 1;
+  JSONNET = 2;
+}
+
+message FileBlob {
+    string file_path = 1;
```
aryx, line 16: Is there a way to have stronger typing for file path? Can you define an alias like
fpath which resolve to a string, but it's still different than a string.

semgrep_runner_v1.proto
```diff
@@ -0,0 +1,38 @@
...
+    // the rule raw content
+    string rule = 1;
+    // how to parse the rule content
+    RuleFormat rule_format = 2;
+    // list of file blobs to receive via SubmitFileBlob service
+    repeated FileBlob files = 3;
+}
+
+message ScanResult {
+    // same as one of the repository_names in the request
```
aryx, line 30: what is this repository_names refered here?

tpetr (approved):
LGTM

aryx (approved):

semgrep_runner_v1.proto
tpetr: Ah I think this was an artifact from our hackweek project. We should probably just delete this field

zyannes: Yup this was a copy/paste error. Will remove!

semgrep_runner_v1.proto
zyannes: This is an option for any clients in the future that _may_ use a Go client. It will ensure that the generated Go client has the proper package/namespace.

semgrep_runner_v1.proto
zyannes: Yes! Added a `FilePath` message type (which simply contains a string called `path`)

semgrep_runner_v1.proto
```diff
@@ -0,0 +1,50 @@
...
+syntax = "proto3";
+
+package semgrep_runner_v1;
+
+option go_package = "semgrep_runner.v1";
+
+import "protos/semgrep_output_v1.proto";
+
+enum RuleFormat {
+  YAML = 0;
```
tpetr, line 19: Nit: let’s follow protobuf best practices with an UNKNOWN = 0 value

tpetr: (maybe in a future PR we could wire up `buf` for proto linting)

zyannes: Good call! Added the `UNSPECIFIED` value and prefix so we don't have to change this later

semgrep_runner_v1.proto
zyannes: Added a comment and included a reference to the server (made this a generic reference to semgrep-proprietary since this is experimental and subject to change)

semgrep_runner_v1.proto
aryx: Looks ugly to have this very Go specific thing in an interface file. There is no such thing for OCaml or python or typescript ...

PR #141, opened 2023-09-26
feat!: new syntax for secrets validators/postprocessors

kopecs:
	  by Semgrep 1.17.0.


akuhlens (approved):
LGTM

PR #143, opened 2023-09-28
add project_metadata.def

jbergler:
`branch` is all sorts of inconsistent and cleaning it up safely is harder than starting to send the new field and progressively migrating uses over to it.

This will also allow us to collect the new values and ensure they're safe before using them.

	  by Semgrep 1.17.0.


jbergler:
Will come back to this later.

PR #149, opened 2023-10-03
feat: new fields in `core_match` for overriding `cli_match` fields

kopecs:
These will be used so that secrets validators can override the severity, message, or metadata for a finding based on runtime behaviour.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -169,7 +169,15 @@ type core_match <python decorator="dataclass(frozen=True)"> = {
...
 type core_match_extra <python decorator="dataclass(frozen=True)"> = {
-  ?message: string option; (* rule.message (?) *)
+  (* These fields generally come from the rule, but may be set here if they're
+   * being overriden for that particular finding. This would currently occur
+   * for rule with a validator for secrets, depending on what the valdiator
+   * might match, but could be expanded in the future. *)
+  ?message: string option;
+  ?metadata: raw_json option;
+  (* LATER: use a variant for severity *)
+  ?severity: string option;
```
aryx, line 179: let's use core_severity here directly instead of those ugly string.
in cli_match_extra we're using a string because we have to, to remain backward compatible, but for new code we should do things in a clean way.

aryx: You can add a (* since 1.44.0 *) comment before metadata and severity.

aryx (approved):

semgrep_output_v1.atd
aryx: Actually regarding core_severity, maybe just rename it "severity", extend it with Inventory and Experiment, so then
we can reuse this type instead of duplicating it everywhere.

aryx, line 179: let's clean this severity thing once and for all!
Let's define it just in the ATD and reuse it in semgrep-core, osemgrep, and pysemgrep.

aryx (changes requested):

semgrep_output_v1.atd
kopecs: Happy to do this, but I'm not sure this solves everything. What's going on with Severity.ml in osemgrep? Should I get rid of that here too?

aryx: I'll take care of Severity.ml

kopecs: @aryx Could you approve changes and the make the required changes for OSS develop, in that case? I'm not sure how you would want to handle it and would rather avoid trying to modify things at the same time.

semgrep_output_v1.atd
```diff
@@ -400,12 +408,14 @@ type incompatible_rule
    warning = something wrong that should be fixed
    info = some special condition worth knowing about
 *)
-type core_severity
-    <ocaml attr="deriving show">
+type severity
+    <ocaml attr="deriving show, eq">
     <python decorator="dataclass(frozen=True)"> = [
   | Error <json name="error">
```
aryx, line 414: I think all of those need to be in uppercase, <json name="ERROR">
according to Cli_json_output.ml and the string_of_severity function

semgrep_output_v1.atd
aryx: yes I can make the change to OSS.

semgrep_output_v1.atd
kopecs: Sure.

semgrep_output_v1.atd
```diff
@@ -209,8 +218,7 @@ type cli_match_extra = {
   (* fields coming from the rule *)
```
aryx, line 218: maybe you can update the comment here as they are not coming directly from the rule now.

semgrep_output_v1.atd
```diff
@@ -400,12 +408,14 @@ type incompatible_rule
    warning = something wrong that should be fixed
    info = some special condition worth knowing about
```
aryx, line 409: would be good to update the comment, to explain the other cases.

aryx (approved):

PR #153, opened 2023-10-06
Separate rule_severity from error_severity

aryx:
test plan:
see related PR in semgrep

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -128,6 +128,46 @@ type rule_id
...
+   Info = some special condition worth knowing about
+   Experiment = guess what
+   Inventory = deprecated: was used for the Code Asset Inventory (CAI) project
+
+   This is used in rules to specify the severity of matches/findings.
+
+   alt: could be called match_severity.
+   coupling: with 'severity' in 'rule_schema_v1.yaml'
+*)
+type rule_severity
```
kopecs, line 143: I would strongly prefer `match_severity` (or `finding_severity` or similar)

semgrep_output_v1.atd
```diff
@@ -439,7 +462,7 @@ type severity
 type cli_error = {
   (* exit code? *)
   code: int;
-  (* TODO: use a variant (Warning | Error | ...) *)
+  (* TODO: use error_severity *)
   level: string;
```
kopecs, line 466: Is there a reason to not do so in this PR?

semgrep_output_v1.py
```diff
@@ -3379,6 +3379,91 @@ def to_json_string(self, **kw: Any) -> str:
         return json.dumps(self.to_json(), **kw)
 
 
+@dataclass(frozen=True)
+class Error_:
```
kopecs, line 3383: Not super relevan to this PR, but this name collision seems annoying to deal with. Is there a reason ATD generates the python like this instead of using [`enum.Enum`](https://docs.python.org/3/library/enum.html)?

semgrep_output_v1.atd
aryx: ok renamed.

semgrep_output_v1.atd
aryx: I am very careful with any change that effect the CLI output. The semgrep-core output, not scared, it's used only internally, but the CLI output, people outside r2c are relying on it so I go slow there and prefer separate PR doing just one thing.

semgrep_output_v1.py
aryx: cc @mjambon on why not using Enum. My guess is that using class generalizes better to support ADTs, so if Error suddently takes some parameter (e.g., a position).

PR #155, opened 2023-10-11
Add Inexistent_file case to skip_reason

aryx:
	  by Semgrep 1.17.0.


aryx:
This is for PA-1344

IagoAbal:
Are you sure this is about this is for PA-1344 "terraform: learn about Terraform and its multifile characteristics" ?

IagoAbal (approved):

PR #159, opened 2023-10-17
Start to use string wrap for uri and other types

aryx:
test plan:
see related PR in semgrep

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -108,11 +107,11 @@ type position
   ~offset: int;
 }
 
-(* File path.
- * TODO: could convert directly to Fpath.t for OCaml with 'string wrap'
- * TODO: could convert directly to Path class of pathlib library for Python
+(* File path. less: could convert directly to Path class of pathlib library for Python
+ * TODO: <ocaml module="ATDStringWrap.Fpath"> but I get compilation
```
aryx, line 111: @mjambon very weird but if you uncomment this, and try to use string wrap for fpath, then I get an ocamlc compilation error for semgrep_output_v1_j.ml. Very bad.

aryx: I've found the bug @mjambon . It's because Fpath.of_string return an option ... so when compiling semgrep_output_v1_j.ml, it is using my ATDStringWrap.ml file which has the wrong Fpath.of_string return type hence the error.
So no bug in ATD! It was my fault.

PR #162, opened 2023-10-18
string wrap part4, ruleid

aryx:
test plan:
see related PR in semgrep

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -13,9 +13,7 @@
  * version of this JSON format. If you need to extend this file, please
  * be careful because you may break consumers of this format (e.g., the
  * Semgrep playground or Semgrep backend or external users of this JSON).
- * See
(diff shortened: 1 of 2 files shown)


semgrep_output_v1.atd
```diff
@@ -106,7 +104,8 @@ type position
 }
 
 (* File path. less: could convert directly to Path class of pathlib library for Python *)
-type fpath <ocaml attr="deriving show"> = string wrap <ocaml module="ATDStringWrap.Fpath">
+type fpath <ocaml attr="deriving show"> =
```
emjin, line 107: Is this change intentional? Looks like just whitespace

emjin (approved):

semgrep_output_v1.atd
aryx: yes just whitespace.

PR #163, opened 2023-10-19
add diff scan related metrics

ihji:
	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -60,10 +60,10 @@ type environment = {
     ?rulesHash <ocaml mutable>: sha256 option;
     (* TODO: should be just boolean *)
     ci <ocaml mutable>: string nullable;
-    (* SHA-256 hash of the commit ID specified via `--baseline_commit` option.
-       If `baseline_commit` option is set, it implies that the scan is operating
+    (* indicates whether `--baseline-commit` option has been specified. When
+       `baseline_commit` option is set, it implies that the scan is operating
        in differential scan mode. Since semgrep 1.46 *)
-    ?baselineCommitHash <ocaml mutable> : sha256 option;
+    ~isDiffScan <python default="False"> <ocaml mutable> : bool;
     ?integrationName <ocaml mutable>: string option;
     ~isAuthenticated <python default="False"> <ocaml mutable>: bool;
   }
@@ -142,7 +142,7 @@ type pro_features = {
 type misc = {
     (* coupling: features is commented a lot in semgrep/PRIVACY.md *)
     features <ocaml mutable>: string list;
-    proFeatures <ocaml mutable>: pro_features;
+    proFeatures <ocaml mutable>: pro_features option;
     ?numFindings <ocaml mutable>: int option;
     ?numIgnored <ocaml mutable>: int option;
     ?ruleHashesWithFindings <ocaml mutable>: (string * int) list <json repr="object"> option;
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -60,6 +60,9 @@ type environment = {
     ?rulesHash <ocaml mutable>: sha256 option;
     (* TODO: should be just boolean *)
     ci <ocaml mutable>: string nullable;
+    (* SHA-256 hash of the commit ID specified via `--baseline_commit` option.
+       If `baseline_commit` option is set, it implies that the scan is operating
+       in differential scan mode. Since semgrep 1.46 *)
     ?baselineCommitHash <ocaml mutable> : sha256 option;
     ?integrationName <ocaml mutable>: string option;
     ~isAuthenticated <python default="False"> <ocaml mutable>: bool;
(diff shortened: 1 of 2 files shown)


semgrep_metrics.atd
```diff
@@ -60,6 +60,7 @@ type environment = {
     ?rulesHash <ocaml mutable>: sha256 option;
     (* TODO: should be just boolean *)
     ci <ocaml mutable>: string nullable;
+    ?baselineCommitHash <ocaml mutable> : sha256 option;
```
aryx, line 63: when adding  new fields, please add a comment before to explain when it was introduced like
(* since semgrep 1.46 *)

aryx: also maybe add a small comment before explaining what it is, in which context this is set.
Most of the other fields do not have comments because this .atd was written first by looking at the code in metrics_.py which didn't have much comment, but for new fields we should try to be better.

semgrep_metrics.atd
```diff
@@ -134,6 +135,7 @@ type misc = {
     features <ocaml mutable>: string list;
     ?numFindings <ocaml mutable>: int option;
     ?numIgnored <ocaml mutable>: int option;
+    ?diffDepth <ocaml mutable>: int option;
```
aryx, line 138: if it's a pro-engine specific thing, maybe we should define another record with pro_features: ... where
we would put this info.

aryx (changes requested):

semgrep_metrics.atd
ihji: Done.

semgrep_metrics.atd
ihji: Done.

semgrep_metrics.atd
```diff
@@ -60,6 +60,10 @@ type environment = {
     ?rulesHash <ocaml mutable>: sha256 option;
     (* TODO: should be just boolean *)
     ci <ocaml mutable>: string nullable;
+    (* SHA-256 hash of the commit ID specified via `--baseline_commit` option.
+       If `baseline_commit` option is set, it implies that the scan is operating
+       in differential scan mode. Since semgrep 1.46 *)
```
aryx, line 65: But do we actually care about the sha256? What we gonna do with this metrics?
Why not just a boolean like is_diff_scan: bool;?

semgrep_metrics.atd
```diff
@@ -129,9 +133,16 @@ type extension = {
...
+type pro_features = {
+    (* The value is determined by the `--diff-depth` option, and it's utilized
+       for inter-file differential scanning. Since semgrep 1.46 *)
+    ?diffDepth <ocaml mutable>: int option;
+}
+
 type misc = {
     (* coupling: features is commented a lot in semgrep/PRIVACY.md *)
     features <ocaml mutable>: string list;
+    proFeatures <ocaml mutable>: pro_features;
```
aryx, line 145: this need to be ?proFeatures ... : pro_features option;
because the metrics backend (which in theory should use this interface file too) need to remain backward compatible and be able to accept metrics generated by old version of Semgrep. See the toplevel comment about protocol update in semgrep_output_v1.atd

aryx (changes requested):

semgrep_metrics.atd
ihji: Updated.

semgrep_metrics.atd
ihji: changed to a boolean type. I added it as a hash just in case we ever wanted to distinguish between runs with different `baseline-commit` values specified, but it's probably not worth it.

semgrep_metrics.atd
```diff
@@ -129,9 +133,16 @@ type extension = {
...
+type pro_features = {
+    (* The value is determined by the `--diff-depth` option, and it's utilized
+       for inter-file differential scanning. Since semgrep 1.46 *)
+    ?diffDepth <ocaml mutable>: int option;
+}
+
 type misc = {
     (* coupling: features is commented a lot in semgrep/PRIVACY.md *)
     features <ocaml mutable>: string list;
+    proFeatures <ocaml mutable>: pro_features option;
```
aryx, line 145: You need the '?' prefix, otherwise if the backend uses this interface file to process the metrics, it will crash
when encountering metrics generated by old version of semgrep.

aryx (changes requested):

semgrep_metrics.atd
ihji: So is it more preferable to use `?proFeatures : pro_features option` rather than `~proFeatures : pro_features`? The linked document seems to use `~z : int` for adding a new optional field.

aryx: I think @mjambon prefers ~, and it seems backward compatible too. I tend to use ? by habits.

ihji: Updated.

semgrep_metrics.atd
```diff
@@ -129,9 +133,16 @@ type extension = {
...
+type pro_features = {
+    (* The value is determined by the `--diff-depth` option, and it's utilized
+       for inter-file differential scanning. Since semgrep 1.46 *)
+    ?diffDepth <ocaml mutable>: int option;
+}
+
 type misc = {
     (* coupling: features is commented a lot in semgrep/PRIVACY.md *)
     features <ocaml mutable>: string list;
+    ?proFeatures <ocaml mutable>: pro_features option;
```
aryx, line 145: would be good to also add (* since 1.46 *) also here

aryx (approved):

semgrep_metrics.atd
ihji: Done. Thanks for the reviews!

PR #172, opened 2023-10-25
run lint in container

jbergler:
Now that the required version of atddiff is available in our base ocaml container, we can use that for much faster runs.


jbergler:
We can try and re-run the job when the docker image has the right version of atddiff in it.

PR #174, opened 2023-10-26
Add config to scan response, and dry_run to scan_metadata

jbergler:
In preparation for refactoring the 3 endpoints needed to create a scan into a single endpoint (and also consolidate the dry-run config endpoint at the same time) this adds types to nicely split up the various components of the create scan + fetch config response.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1039,29 +1040,35 @@ type scan_request = {
...
+type scan_configuration = {
+    enabled_products: product list;
+    rules: raw_json; (* can we type this better *)
+    ~ignored_files: string list;
+    ~triage_ignored_syntactic_ids: string list;
+    ~triage_ignored_match_based_ids: string list;
 }
-*)
 
-(* TODO? seems used only for dry_run and passed urlencoded to /scans/config?xxx
```
jbergler, line 1054: removing this since my goal is to remove this endpoint entirely in the next 1-2 versions of the cli

semgrep_output_v1.atd
```diff
@@ -1039,29 +1040,35 @@ type scan_request = {
...
-type scan_config_request = {
-    dry_run: bool;
-    (* project_metadata.repository *)
-    repo_name: string;
-    (* TODO: deprecate, prefer scan_request.scan_metadata.requested_products *)
-    sca: bool;
-    full_scan: bool;
-    semgrep_version : version;
+(* settings for the cli *)
```
jbergler, line 1065: this is where i think generic config options for the cli should go

semgrep_output_v1.atd
```diff
@@ -909,8 +909,8 @@ type tag = string
 
 (* Response by the backend to the CLI to the POST /scans/<scan_id>/config *)
 type scan_config = {
```
aryx, line 911: So I guess we can delete this type then in favor of your new 'scan_configuration'

semgrep_output_v1.atd
```diff
@@ -909,8 +909,8 @@ type tag = string
...
+    (* TODO: remove deployment_id + deployment_name from here, instead this will be in scan_response *)
```
aryx, line 912: you mean scan_info?

semgrep_output_v1.atd
```diff
@@ -1039,29 +1040,35 @@ type scan_request = {
...
+type scan_info = {
+    id: string; (* the scan id *)
+    deployment_id: int;
+    deployment_name: string;
+}
+
+(* config specific to the scan, eg *)
+type scan_configuration = {
+    enabled_products: product list;
+    rules: raw_json; (* can we type this better *)
```
aryx, line 1059: actually here we're sending a string (that should be interpreted as JSON, but this is still a string)

semgrep_output_v1.atd
aryx: Hmm, I think this is used for the 'semgrep scan --config policy'
that is we're not really starting a CI scan (and upload findings), but we want to fetch the rules from the ruleboard

semgrep_output_v1.atd
jbergler: not in the new version of the endpoint

semgrep_output_v1.atd
jbergler: Yes, if it's not used at all (which seems likely) then we can remove this, I'd vote for doing this in a separate PR though.

semgrep_output_v1.atd
jbergler: Yeah bit of both - in `scan_response.scan_info`

semgrep_output_v1.atd
jbergler: Yeah, I think we should update the CLI to use the same endpoint as I'm adding for CI at the moment. But probably in a separate commit.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+semgrep_output_v1* linguist-generated=true
```
aryx, line 1: very nice!
This is useful.

semgrep_output_v1.atd
```diff
@@ -1113,29 +1114,35 @@ type scan_request = {
...
+(* meta info about the scan *)
+type scan_info = {
+    id: int; (* the scan id *)
+    deployment_id: int;
+    deployment_name: string;
+}
+
+(* config specific to the scan, eg *)
+type scan_configuration = {
+    enabled_products: product list;
```
aryx, line 1132: But this will break old CLI right? They are expecting this enabled_products thing
under scan:, not under scan_config: in the scan_response.

semgrep_output_v1.atd
```diff
@@ -1113,29 +1114,35 @@ type scan_request = {
...
-type scan_config_request = {
-    dry_run: bool;
-    (* project_metadata.repository *)
-    repo_name: string;
-    (* TODO: deprecate, prefer scan_request.scan_metadata.requested_products *)
-    sca: bool;
-    full_scan: bool;
-    semgrep_version : version;
+(* settings for the cli *)
+type engine_configuration = {
```
aryx, line 1140: I'm not sure about the split between scan_configuration and engine_configuration.
Maybe better to merge them. Those are both engine configuration; What is an engine config vs scan config?

semgrep_output_v1.atd
```diff
@@ -1113,29 +1114,35 @@ type scan_request = {
 }
 
 (* Response by the backend to the CLI to the POST /scans *)
-(* TODO
 type scan_response = {
-    scan: scan_response_bis;
+    scan: scan_info;
+    scan_config: scan_configuration;
```
aryx, line 1119: would be good to introduce here the new and more general 'ci_config_from_cloud'
that hopefully the Frontend and backend could start to use to send scan config info.

semgrep_output_v1.atd
aryx: In fact I started with zach this new 'ci_config_from_cloud' that tries to be the more general and new way
to send scan config configs from the cloud to the CLI. Wou;ld be great to deprecate
this engine_configuration and rely on ci_config_from_cloud instead.

aryx (approved):

semgrep_output_v1.atd
jbergler: We should probably chat about this more. I know you just added `ci_config_from_cloud` but I really want to get rid of it and replace it with this.

The intention behind the design here is that `scan_config` encapsulates all the state that changes scan to scan, and is tightly coupled with policy/rule selection for a particular repo. This gets built by calling out to various products `generate_config` functions and assembled. In the future, we may want to handle this differently in the backend.

`engine_config` on the other hand is relatively static, gets generated per deployment, with some possibility for per repo overrides in the future, but is ultimately product agnostic. 

fyi @zyannes

jbergler: Also, I should mention that "ci" is just one kind of scan and I'm trying to work towards unifying the APIs used by `semgrep scan` and `semgrep ci` so that's another reason why I don't want CI in the name of the type.

semgrep_output_v1.atd
zyannes, line 1119: super-nit: prefixing the fields with `scan` feels redundant. What do you think about naming them `info` and `config` respectively?

semgrep_output_v1.atd
```diff
@@ -1113,29 +1114,35 @@ type scan_request = {
...
+    engine_config: engine_configuration;
 }
-type scan_response_bis = {
-    (* the scan id *)
-    id: string;
-    (* since 1.37.0 *)
-    ?enabled_products: product list;
+
+(* meta info about the scan *)
+type scan_info = {
```
zyannes, line 1124: What about splitting this into separate types: `scan_id` and `scan_deployment`? Then the scan_response can simply have 
```
id: scan_id;
deployment: scan_deployment;
```

zyannes: Yeah I suppose you're right. Having this type defined means we can add additional scan info metadata later if needed.

semgrep_output_v1.atd
jbergler: made a couple of tweaks.

semgrep_output_v1.atd
aryx: actually we want to differentiate more scan and ci with @milanwilliams and @zzeleznick  so I'm not sure 'semgrep scan' should  do anything with the backend. It should be mostly for local use.

aryx: But yes we should chat more about ci_config_from_cloud and who does what.

PR #176, opened 2023-10-27
Type the deployment_config response from the backend to the CLI and a few other responses

aryx:
test plan:
see related PR in semgrep

	  by Semgrep 1.17.0.


aryx:
@mjambon I think atddiff is wrong here. This PRs introduces new types, so we should
not complain about it.

semgrep_output_v1.atd
```diff
@@ -1230,8 +1249,21 @@ type ci_scan_dependencies = (string * found_dependency list) list
     <python repr="dict">
     <ts repr="map">
 
+
+
+
+(* Response by the backend to the CLI to the POST /results *)
```
jbergler, line 1255: These changes are unrelated to the change in the title/pr description, can they be broken out into a separate PR?

semgrep_output_v1.atd
```diff
@@ -898,6 +894,29 @@ type dependency_parser_error = {
 (* CI Configuration *)
 (* ----------------------------- *)
 
+(* Response by the backend to the CLI to the POST deployments/current *)
+type deployment_config <ocaml attr="deriving show"> = {
```
jbergler, line 898: Nit picky, but "config" is the wrong word to use here IMO - config implies something else, and I'm trying to split these concepts up in a separate PR.

My preference would be to not type this endpoint at all since it's not dedicated to the CLI and therefore will never be typed with ATD in the backend. 

In a separate PR that I'm working on backend changes for I actually stop calling this endpoint anyway.

aryx: I'll be happy to remove this type from the ATD, but right now it's still used in osemgrep so I need it.

semgrep_output_v1.atd
aryx: yes ... I've updated the title at least, but this commit is already referenced in semgrep
and used for semgrep-interfaces so I'll merge this one.

aryx:
I think atddiff is wrong here @mjambon so I'll force merge.

PR #177, opened 2023-10-30
Add comment about paired type

akuhlens:
	  by Semgrep 1.17.0.


akuhlens:
I rebased on top of what is pointed to in the semgrep repo. So that isn't pointing to the main branch here.

aryx (approved):

semgrep_output_v1.atd
```diff
@@ -177,6 +177,7 @@ type engine_kind
 type rule_id_and_engine_kind <python decorator="dataclass(frozen=True)"> =
   (rule_id * engine_kind)
 
+(* Need to maintain pairity between this and Inputs_to_core.product. *)
```
aryx, line 180: It's worth than that because @ajbt200128 I think recently added a type product in Rule_filtering.ml
Would be good to consolidate and have just one ...

PR #180, opened 2023-11-03
feat: add additional internal pattern combinator

kopecs:
Allows `semgrep-internal-patterns-allow-disjoint` to appear everywhere (other than taint) where a `patterns` may appear. See notion for semantics.

	  by Semgrep 1.17.0.


aryx (commented):
What is that? Is there some doc about it and about its motivation?

kopecs:
Yep - see https://www.notion.so/semgrep/semgrep-internal-patterns-allow-disjoint-bd56b763e470494bba9213523c45a5cf?pvs=4

rule_schema_v1.yaml
```diff
@@ -212,6 +215,7 @@ $defs:
           - required: [ pattern-either ]
           - required: [ pattern-not ]
           - required: [ pattern-inside ]
+          - required: [ anywhere ]
```
aryx, line 218: I guess pattern-anywhere here? to be consistent with the other?

aryx (approved):

rule_schema_v1.yaml
kopecs: I don't feel strongly; if anything I would rather just make it `semgrep-internal-pattern-anywhere` and then just remove that prefix & document this when we look to transition to the "new" syntax at some future point.

More broadly, consistency-wise, it is already inconsistent with how `inside`/`pattern-inside` works in the old syntax since it takes any pattern rather than just a leaf (whereas `pattern-inside` only lets you do a leaf pattern, in the sense of a string rather than a formula)

aryx: True, I forgot about that.

aryx: go for semgrep-internal-pattern-anwhere in old syntax and just 'anywhere' in new syntax, to give another incentive to use the new syntax :)

semgrep_output_v1.atd
```diff
@@ -685,6 +685,7 @@ type matching_operation <ocaml attr="deriving show { with_path = false}"> = [
   | And
   | Or
   | Inside
+  | Anywhere
```
aryx, line 688: we usually try to indicate when something was introduced, like (* since semgrep 1.49 *)

kopecs: I'll put up another PR to add a note.

PR #185, opened 2023-11-07
add multi-type syntax to metavariable-type rule

ihji:
	  by Semgrep 1.17.0.


aryx (approved):
Is this supporting the new syntax too?

ihji:
> Is this supporting the new syntax too?

it supports both rule v1 and v2.

PR #189, opened 2023-11-09
Update types to return the languages for which we tried to run the pro engine

emjin:
This will make it easier for us to track the impact of new interfile languages (usage, error rate, performance).

	  by Semgrep 1.17.0.


semgrep_metrics.atd
```diff
@@ -166,4 +166,5 @@ type misc = {
     ?ruleHashesWithFindings <ocaml mutable>: (string * int) list <json repr="object"> option;
     (* TODO: should be OSS | Pro, see semgrep_output_v1.atd engine_kind type *)
     ~engineRequested <python default="'OSS'"> <ocaml mutable>: string;
+    ?requestedInterfileLanguages: string list option;
```
aryx, line 169: add a (* since Semgrep 1.49.0 *) comment please

semgrep_output_v1.atd
```diff
@@ -764,6 +764,12 @@ type cli_output_extra = {
 
     (* EXPERIMENTAL: since: 1.37.0 *)
     ~skipped_rules: skipped_rule list;
+
+    (* Reporting just the requested engine isn't granular enough. We want to know
+       what languages had rules that invoked interfile. This is particularly important 
+       for tracking the performance impact of new interfile languages *)
+    (* EXPERIMENTAL: since 1.49.0 *)
+    ?requested_interfile_languages: string list option;
```
aryx, line 772: 'requested' then is maybe not a good prefix. It makes it sound like the user did --interfile java. maybe 'effective_...' ? also maybe move the field just below ?engine_requested, so the relation between the 2 is clearer (the order of the fields does not matter for backward compatibility, don't worry). Maybe interfile_languages_used ?

aryx (approved):

PR #202, opened 2023-12-13
feat: more granular pro engine config metrics

kopecs:
Creates a new metrics type which is closely aligned with osemgrep's Engine_type.t which allows for more precise tracking of the engine feature matrix. This is needed compared to a simple enumeration since we have more orthogonal features than previously. For instance, we can now perform analysis on at least two axes (within pro):
    (1) secrets valdiation [enabled <-> disabled];
    (2) dataflow [none <-> ... <-> interfile]
Creating a new enum entry for each possible permutation is not futureproof (and is ugly), so we need a more complex type here.

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -156,6 +156,33 @@ type pro_features = {
     ?diffDepth <ocaml mutable>: int option;
 }
 
+type analysis_type <ocaml attr="deriving show"> = [ 
+    | Intraprocedural 
+    | Interprocedural 
+    | Interfile
+]
+
+type secrets_config 
+     <ocaml attr="deriving show"> = {
+  permitted_origins: string;
+}
+
+type pro_config 
+     <ocaml attr="deriving show"> = {
+  analysis_type: analysis_type;
+  ?secrets_config: secrets_config option;
+  pro_langs: bool;
+}
+
+(* Since v1.54.0 *)
+type engine_config
+     <ocaml attr="deriving show">
+     <python decorator="dataclass(frozen=True)"> = [
+  | OSS
+  | PRO of pro_config
+]
+
+
 type misc = {
     (* coupling: features is commented a lot in semgrep/PRIVACY.md *)
     features <ocaml mutable>: string list;
@@ -165,7 +192,10 @@ type misc = {
     ?numIgnored <ocaml mutable>: int option;
     ?ruleHashesWithFindings <ocaml mutable>: (string * int) list <json repr="object"> option;
     (* TODO: should be OSS | Pro, see semgrep_output_v1.atd engine_kind type *)
+    (* Still needed with engineConfig? *)
     ~engineRequested <python default="'OSS'"> <ocaml mutable>: string;
+    (* Since Semgrep 1.54.0 *)
+    ?engineConfig <ocaml mutable>: engine_config option;
     (* Since Semgrep 1.49.0 *)
     ?interfileLanguagesUsed: string list option;
   }
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -162,26 +162,28 @@ type analysis_type <ocaml attr="deriving show"> = [
     | Interfile
 ]
 
+type code_config <ocaml attr="deriving show"> = unit
+
+type secrets_origin <ocaml attr="deriving show"> = [ Any | Semgrep ]
 type secrets_config 
      <ocaml attr="deriving show"> = {
-  permitted_origins: string;
+  permitted_origins: secrets_origin;
 }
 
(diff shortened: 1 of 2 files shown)


semgrep_metrics.atd
```diff
@@ -156,6 +156,33 @@ type pro_features = {
...
+     <ocaml attr="deriving show"> = {
+  permitted_origins: string;
+}
+
+type pro_config 
+     <ocaml attr="deriving show"> = {
+  analysis_type: analysis_type;
+  ?secrets_config: secrets_config option;
+  pro_langs: bool;
+}
```
emjin, line 175: I'd like this to clearly be our central place where we report the configuration of products that we ran. e.g.
```
type products_config = {
    code: code_config option;
    secrets: secrets_config option;
    supply_chain: supply_chain_config option;
}
```

where `code_config` would be `analysis_type` and `supply_chain_config` can be `()` for now if there's no config. I want to make it really easy to answer the question "what products did x run with?"

The one thing is I realize that `analysis_type` affects both Code and Secrets. Frankly, I think it shouldn't, and we should have them configured separately. Because of that preference, for now, I would make analysis_type separately a part of secrets_config.

semgrep_metrics.atd
```diff
@@ -165,7 +192,10 @@ type misc = {
     ?numIgnored <ocaml mutable>: int option;
     ?ruleHashesWithFindings <ocaml mutable>: (string * int) list <json repr="object"> option;
     (* TODO: should be OSS | Pro, see semgrep_output_v1.atd engine_kind type *)
+    (* Still needed with engineConfig? *)
```
emjin, line 195: Let's say something like "TODO deprecate this by 2/14/2023 and rely on pro_config"

semgrep_metrics.atd
```diff
@@ -156,6 +156,33 @@ type pro_features = {
...
+  pro_langs: bool;
+}
+
+(* Since v1.54.0 *)
+type engine_config
+     <ocaml attr="deriving show">
+     <python decorator="dataclass(frozen=True)"> = [
+  | OSS
+  | PRO of pro_config
+]
```
emjin, line 183: In the world where we just have `product_config`, I wouldn't have this `engine_config` idea. 

I think this PR really reveals the existing problems with the paradigm of "engineRequested", which made sense when we only had one product where this was relevant but now feels weird. It feels strange to me to have the secrets configuration be under the `pro_config` and muddles even further what "pro" means.

emjin (changes requested):
Normally I would approve a PR like this and say "it's a strict improvement over the existing state, you don't have to fix the tech debt". For metrics changes, though, having a different intermediate state affects people's queries. I think it's worth being forward looking. I'd like this change to help bring us to a more straightforward product-oriented world.

semgrep_metrics.atd
kopecs: > where code_config would be analysis_type and supply_chain_config can be () for now if there's no config. I want to make it really easy to answer the question "what products did x run with?"

Sure. I can add code and supply chain here. I didn't since there were not obvious configuration options to me which needed to be associated with code and from asking around it I was told there were not relevant options for supply chain to add here.

> The one thing is I realize that analysis_type affects both Code and Secrets. Frankly, I think it shouldn't, and we should have them configured separately.

Like I mentioned, I disagree. I think it just makes running rules more complicated if we allow mixing things like this. If we're already doing the naming resolution, e.g., we should just have everything benefit from that instead of restricting it and having to ensure it isn't used for only some rules.

semgrep_metrics.atd
kopecs: Sure.

semgrep_metrics.atd
kopecs: We could just make `engine_config` be `pro_config option` (and rename somewhat) if there are objections to calling this "pro_xxx".

semgrep_metrics.atd
emjin: > Like I mentioned, I disagree. I think it just makes running rules more complicated if we allow mixing things like this. If we're already doing the naming resolution, e.g., we should just have everything benefit from that instead of restricting it and having to ensure it isn't used for only some rules.

Reflected on this more. I think what I really dislike right now is that it isn't clear that analysis_type affects both when you go to the toggle in the app. I'd be fine with having a `global_config` for options that affect all products

emjin (approved):
This looks a lot better to me. I think it's still worth asking Milan or Raghav if they have any suggestions from the queryer perspective, but I think it makes sense

kopecs:
Lovely; github's rebase button doesn't sign.

PR #207, opened 2024-01-06
Tentative fixes for the atddiff check

mjambon:
I guessed possible problems from CI error logs and fixed the following:

* The GHA script may be running with Bash without `-o pipefail` which would cause the check command within a pipe to not make the CI check fail. I don't know if this problem really happened but specifying `-o pipefail` regardless of how GHA invokes Bash removes doubt.
* If the JSON containing the minimum version ID can't be fetched, it now results in a clear error message rather than an invalid tag `v` and strange error later. I saw this error in https://github.com/semgrep/semgrep-interfaces/actions/runs/7036193485/job/19148273385?pr=199

test plan:
* check out this branch.
* add a dummy required field to a record such as adding `floobz: int;` to the `location` type definition.
* commit this change
* run the script with `./scripts/check-backwards-compatibility`
* check the exit status with `echo $?`; it should be nonzero. Ideally, it should be 3, but an apparent bug in `git difftool` causes it to be 128.
* undo the dummy commit

	  by Semgrep 1.17.0.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -40,7 +40,7 @@ jobs:
         atddiff --version
 
         # run the checks
-        echo -ne 'Backwards compatability summary:\n\n```' > summary-00-header.txt
+        echo -ne 'Backwards compatibility summary:\n\n```' > summary-00-header.txt
         echo '```' >> summary-20-footer.txt
 
         # fail if check command fails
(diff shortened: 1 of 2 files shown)


jbergler (approved):

mjambon:
Closed as duplicate of #207.

PR #210, opened 2024-01-08
feat(metrics): add findings by product breakdown

amchiclet:
Helps PA-3312.

This breakdown will allow us to calculate the number of findings per bytes of code scanned per product to understand users better.

	  by Semgrep 1.17.0.


emjin (approved):
👀

PR #214, opened 2024-01-19
add go_package to the protobuf generator

tpetr:
Setting this makes it considerably easier for golang users to use the protobuf generated code.

	  by Semgrep 1.17.0.


tpetr:
Nevermind, this is more complicated than expected. I can fix this at protobuf generation time in semgrep-app

PR #216, opened 2024-01-23
feat!: remove deprecated postprocessor syntax

kopecs:
Removes the `mode` setting for the old postprocessor syntax we replaced 3-4 months ago. This syntax is undocumented and unused internally; externally it was never part of a documented public API.

BREAKING CHANGE: This means we may now reject rules with `mode: semgrep_internal_postprocessor`

	  by Semgrep 1.17.0.

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -827,8 +827,6 @@ properties:
             # EXPERIMENTAL
             - join
             - extract
-            # EXPERIMENTAL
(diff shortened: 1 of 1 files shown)


aryx (approved):
Note that now there is also a rule_schema_v2.atd for the Semgrep syntax, but
it probably already didn't cover this old syntax so we should be fine.

kopecs:
Yes, I believe it only has the new/current syntax.

PR #218, opened 2024-01-25
add: os and isjs environment metrics

ajbt200128:
See: https://github.com/semgrep/semgrep/pull/9670

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -66,8 +66,10 @@ type payload = {
 type environment = {
     (* semgrep version *)
     version <ocaml mutable>: string;
-    os: string;
-    isJS <ocaml mutable>: bool;
+    (* Since Semgrep 1.59 *)
+    os: string; (* e.g. linux *)
+    (* Since Semgrep 1.59 *)
+    isTranspiledJS <ocaml mutable>: bool;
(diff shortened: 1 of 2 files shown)


semgrep_metrics.atd
```diff
@@ -66,6 +66,8 @@ type payload = {
 type environment = {
     (* semgrep version *)
     version <ocaml mutable>: string;
+    os: string;
```
aryx, line 69: How will the backend react to those new fields? I guess we're not using yet semgrep_metrics.atd on the backend side
so I guess it's fine, but better to think about it.

Maybe you can also add a comment before like (* since Semgrep 1.59 *).

also isJS is a bit vague, so would be better to explain what it is, and for os maybe give an example (* e.g., "linux" *)

aryx (approved):

semgrep_metrics.atd
ajbt200128: From what I've seen the backend usually manually accesses fields i.e. data["version"] so new fields should be ok, I couldn't find a breaking instance, but there may be a place I missed where this data is reported

ajbt200128: I'll add comments, not sure what to call isJS, maybe isTranspiledJS or isJSOO?

PR #219, opened 2024-01-30
feat(metrics): add findings_by_product field for CI scan stats payload

zzeleznick:
## Description
This PR supports adding the number of findings per product for our CI runs. Follow-on to #210. 

**Addresses**: GROW-104 

### Test Plan
	  by Semgrep 1.17.0.


amchiclet (approved):

semgrep_output_v1.atd
```diff
@@ -1351,6 +1351,18 @@ type ci_scan_complete_stats = {
    * used in this file.
    *)
   ?engine_requested: string option;
+  (* Mirrors numFindingsByProduct in metrics.py
+   * See PA-3312 and GROW-104.
+   *
+   * NOTE: As of 1.56.0 the string used as the mapping key is
+   *  currently a human-readable product name (i.e. code)
```
aryx, line 1358: would be better to use an actual enum ... We actually have a type product.
but yes maybe you can't use an enum for the key of a dictionnary with ATD.

PR #220, opened 2024-01-31
feat(secrets): Add field for historical commit data

kopecs:
This is useful for secrets historical scans, which may result in findings at commits other than the current HEAD. In these cases we need to report at least the git commit associated with the finding for identification purposes.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -914,6 +920,14 @@ type dependency_parser_error = {
   ?text: string option;
 }
 
+(*****************************************************************************)
+(* Semgrep Secrets *)
+(*****************************************************************************)
+
+type historical_info = {
+  git_commit: sha1;
+  git_commit_timestamp: (* ISO 8601 datetime *) string;
```
aryx, line 929: there is a datetime type at the beginning of this file. Would be good to use a consistent type and format for dates.

semgrep_output_v1.atd
```diff
@@ -914,6 +920,14 @@ type dependency_parser_error = {
   ?text: string option;
 }
 
+(*****************************************************************************)
+(* Semgrep Secrets *)
+(*****************************************************************************)
+
+type historical_info = {
+  git_commit: sha1;
```
aryx, line 928: maybe add a comment explaining what is this about; why we suddently need to commit associated with a match.

aryx (approved):

semgrep_output_v1.atd
kopecs: Sure.

semgrep_output_v1.atd
kopecs: Sure, I can change this and use `string wrap` for the alias I think.

kopecs: It looks like we somewhat use the existing `ATD_string_wrap.Datetime` but only internally for the purpose of conversion to a string in osemgrep output. As-is, that module seems to use git's "ISO 8601-like" format without a timezone. I think it would be better if we used a more standard format (e.g., RFC 3339), and use a dedicated date/time library (we already use `Timedesc` for some other purposes) for handling that.

I've changed `datetime` here to use the wrapper module and plan on updating the code in semgrep to use `Timedesc` in the wrapper module, have it output RFC 3339 formatted datetimes and update the one existing usage of `ATD_string_wrap.Datetime` to no longer use that module but have the same behaviour with the idea being we can update it to be the same in the future. I hesitate to change the format of the existing field along with this unless I can get confirmation that the existing handling of timestamps would allow us to send a timezone (which would seem reasonably likely to me).

ajbt200128 (approved):
nice

semgrep_output_v1.atd
kopecs: After some investigation it appears we only ever use https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat in the app, which should also handle this, so we ought to be covered there.

PR #223, opened 2024-02-09
Add an option for adding a separate display name for repositories.

emjin:
	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1110,6 +1110,10 @@ type project_metadata = {
     ?repo_id: string option;
     (* a.k.a repository owner id *)
     ?org_id: string option;
+    
+    (* Users can set a different name for display and for PR comments.
+       This allows monorepos to be scanned as separate projects. *)
+    repo_display_name;
```
aryx, line 1116: need ?, otherwise breaks backward compatibility (as reported correctly by @mjambon 's tool atddiff).

aryx (changes requested):

semgrep_output_v1.atd
emjin: Great tool 🤩

semgrep_output_v1.atd
```diff
@@ -1111,6 +1111,10 @@ type project_metadata = {
     ?repo_id: string option;
     (* a.k.a repository owner id *)
     ?org_id: string option;
+    
+    (* Users can set a different name for display and for PR comments.
```
aryx, line 1115: Note that I introduced a ci_config_from_cloud type in this file, that has a dirs_config: field
especially designed for monorepos, to allow to have different configuration per directory.
It's not used yet by the backend, but I think we should start using it, to become monorepo friendly.

aryx (approved):

semgrep_output_v1.atd
emjin: @chmccreery I know your team has other priorities but I am pretty concerned that the env var method of configuring monorepos won't be ergonomic, which would still hamper adoption. Would love to sync on what we could or couldn't accomplish quickly with a cross-team project here.

aryx: Yes, I think we should go towards the ci_config_from_cloud data structure I defined above.
That way we get all the info from the cloud in a uniform way.

aryx: (and typed way)

chmccreery: Yeah, I am in no way wedded to env variables, though I think they are useful for allowing customization/overrides. If you wanted to auto-detect this variable for monorepos, that seems fine as well to me! Maybe I'm missing the context on what would be needed from the backend in this case.

emjin: I think (@aryx correct me if I'm wrong) that the suggestion is to allow people to select their sub repos in the app. The proposed config is:

```
type ci_config_from_cloud = {
    repo_config: ci_config;
    ?org_config: ci_config option;
    (* for monorepos, to be "monorepo-friendly" like they say in Ruff *)
    ?dirs_config: (fpath * ci_config) list option;
}
```

where the app would send us a list of the directories to scan individually. I'm not sure if I think that's the way that makes the most sense, but I do think for highest adoption we will want people to be able to configure sub repos in the app, and if so it would be ideal for the display name to be sent through a typed interface rather than setting an env var.

emjin: Separately I have been thinking about whether it would make sense to autodetect the display name. We could say that when we detect that we're in a subfolder, the display name is `REPO_NAME.subfolder_name`. This should be pretty easy on my end; the only question is whether there's some common use case I'm overlooking that this would block.

chmccreery: I can imagine `REPO_NAME.subfolder_name` being non-unique unless you use a full path for the `subfolder_name`. But I think the autodetection would make sense!

If app sent a list of directories to scan individually, would the user only have to trigger `semgrep ci` once from the root, but then it would spawn a bunch of different scans that cd into their respective directories? Or would the user still presumably scan the subdirectories separately and at different times?

aryx: the GHA would just run semgrep ci at the root, and semgrep ci will fetch this repo config from our backend, which would return this ci_config_from_cloud interface emma copy-pasted above that indicates how each directory should be managed.

aryx: This is why it's important we agree on the interface of this ci_config_from_cloud and its dirs_config: field, so 'semgrep ci' can get the appropriate information from the backend.

aryx: Once we agree on the interface (in semgrep_output_v1.atd), then we can independently implement things.

chmccreery: Feel free to propose an interface, I'm happy to review it! If it requires much of a UI/design effort, it may not happen for awhile (which is why the env variable is a nice thing to have in the interim).

aryx: I have, this is the ci_conflig_from_cloud that @emjin copy pasted above. Here it is again:
```
(* Semgrep config from the WebApp
 * TODO: not created yet by backend, and not used yet in the CLI
 *)
type ci_config_from_cloud = {
    repo_config: ci_config;
    ?org_config: ci_config option;
    (* for monorepos, to be "monorepo-friendly" like they say in Ruff *)
    ?dirs_config: (fpath * ci_config) list option;
}

(* Note that we should use very simple types below for the configuration
 * of Semgrep: booleans or small enums. No int, as people often don't
 * understand how to set values. For example even if we documented
 * very well the --timeout option in Semgrep, people still didn't
 * know which value to use.
 *
 * LATER: the type below could be used for the automatic generation of UI code
 * in the WebApp for user to setup the CI config with UI widgets. We would need
 * probably ATD extension to express validators, docstring, etc (Jonas's idea).
 *
 *)
type ci_config = {
   (* to override environment variables, as lots of the configuration of
    * 'semgrep ci' comes from environment variables (e.g., SEMGREP_REPO_URL)
    *)
    env: ci_env;
    enabled_products: product list;
    (* glob patterns *)
    ignored_files: string list;
    (* other features *)
    autofix: bool;
    (* TODO?
     *  - feature_rollout (hidden from users)
     *  - feature_opt_in (set by users)
     *  - deepsemgrep? but should we use product instead?
     *  - dependency_query??
     *  - triage_xxx?
     *)
}

type ci_env = (string * string) list
  <json repr="object">
  <python repr="dict">
  <ts repr="map">

```


For context we also have:
```
(* Content of the .semgrepconfig.yml in the repository.
 *
 * This config allows to configure Semgrep per repo, e.g., to store
 * a category/tag like "webapp" in a repo so that the Semgrep WebApp can
 * return a set of relevant rules automatically for this repo in scan_config
 * later when given this ci_config_from_repo in the scan_request.
 *)
type ci_config_from_repo = {
    (* version of the .semgrepconfig.yml format. "V1" right now (useful?) *)
    ~version <python default="Version('v1')"> <ts default="'v1'">: version;
    ?tags: tag list option;
}
(* ?? ex? *)
type tag = string

```

aryx: More context:
```
(* Response by the backend to the CLI to the POST /scans/<scan_id>/config *)
type scan_config = {
    (* TODO: remove deployment_id + deployment_name from here, instead this will be in scan_response.scan_info *)
    deployment_id: int;
    deployment_name: string;
    (* ex: "audit", "comment", "block" TODO use enum? TODO: seems dead *)
    policy_names: string list;
    (* rules raw content in JSON format (but still sent as a string) *)
    rule_config: string;
    (* since 1.47.0 *)
    ?ci_config_from_cloud: ci_config_from_cloud option;

    (* Deprecated: should rely on ci_config_from_cloud instead *)
    ~autofix: bool;
    ~deepsemgrep: bool;
    ~dependency_query: bool;
    ~triage_ignored_syntactic_ids: string list;
    ~triage_ignored_match_based_ids: string list;
    (* glob patterns *)
    ~ignored_files: string list;
    (* since 1.37.0 *)
    ?enabled_products: product list option;
  }

```
But really you can just read the end of the semgrep_output_v1.atd which contains the types to communicate with the backend.

chmccreery: For this
```
type ci_config_from_cloud = {
    repo_config: ci_config;
    ?org_config: ci_config option;
    (* for monorepos, to be "monorepo-friendly" like they say in Ruff *)
    ?dirs_config: (fpath * ci_config) list option;
}
```

What is the benefit of having the app send all of this, instead of having the app do all of the override logic and just send the final values to the CLI? For example, it seems to be that it would be simpler and therefore better to send one, synthesized list of `ignored` files to the CLI and abstract away whether they were configured at the org, repo, or subdirectory (for a monorepo) level. 

I was imagining that the CLI could send to the app the token (for deployment config), repo_display_name (repository config), and current subdirectory (for dirs config) and the app could send back just a single `ci_config` instead of multiple.

chmccreery: I am also biased toward putting this type of logic in the backend because when we make changes (e.g. add support for monorepos there) it wouldn't have to involve a change to the interface, which always requires more complexity in the app as we have to support multiple versions of the CLI.

aryx: maybe we can have a meeting, but I think we want to reduce the number of roundtrips between the CLI and the backend, so the idea is that for monorepos, the CI would just run 'semgrep ci', and semgrep ci itself would do the entire scan and use different config depending on the directory it's in, so 'semgrep ci' at the beginning would fetch this config from the backend.

aryx: Moreover just regarding the semgrep ignored, it's actually not trivial to compute given a set of possibly deeply nested .semgrepignore how to ignore files; Martin actually had to implement this stuff in osemgrep, which actually goes further of what pysemgrep does (which just handle a toplevel .semgrepignore), and imitate what gitignore is doing, and it's not trivial.

emjin: > the CI would just run 'semgrep ci'

I'm not sure this is what users want. I'm not sure what they want (I haven't spoken to people about this) but I imagine that if your entire codebase is one monorepo you could want to let your teams run semgrep separately on each repo, the way we do for semgrep vs semgrep-app. Maybe worth speaking to Milan first to figure out the preferred use.

emjin: The other concern with one `semgrep ci` run is that I'm worried it doesn't actually help us solve the main problem, which is CI timing out on long semgrep scans.

emjin: This random github thread is getting long; @chmccreery do you have a preferred place to discuss something like this? Maybe an "RFC" ticket on Findings?

aryx: maybe ... milan knows maybe better the actual use case. But anyway, I'd rather not (ab)use environment variables to store useful information; it's not typed, it's not explicit, it's fragile. I'd rather have actual fields in semgrep_output_v1.atd

PR #225, opened 2024-02-12
feat(secrets): add engine_config setting for historical scans

kopecs:
This allows for the app to send configuration for historical scans when a scan requests the configuration from the app.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -931,6 +931,10 @@ type historical_info = {
    * secrets which are buried in the git history which we wouldn't find at HEAD
    *)
   git_commit: sha1;
+  (* Git blob at which the finding is present. Sent in addition to the commit
+   * since some SCMs have permalinks which use the blob sha, so this information
+   * is useful when generating links back to the SCM. *)
+  git_blob: sha1;
```
salolivares, line 937: Should we also keep git_blob as optional to make it backwards compatible?

kopecs: Yeah, I was just writing a comment about this. The record in which this currently appears is not yet used, so there isn't a real compatibility concern and I am skeptical the added complexity of making this optional at the usages is better than not.

salolivares: Ahh, gotcha. If this is part of a record that is unused, then LGTM.

semgrep_output_v1.atd
```diff
@@ -241,7 +241,7 @@ type core_match_extra <python decorator="dataclass(frozen=True)"> = {
    * Added in semgrep 1.37.0 *)
   ?validation_state : validation_state option;
   (* EXPERIMENTAL: added by secrets post-processing & historical scanning code.
-   * Since 1.60.0. *)
+   * Since 1.62.0. *)
```
emjin, line 244: I assumed this comment was attached to validation_state?

emjin (approved):
Seems reasonable to me

semgrep_output_v1.atd
kopecs: This is for `historical_info`. We haven't sent it yet so I would just be updating this comment here.

PR #226, opened 2024-02-24
Add an engine flavor type to report findings that can only come from pro

emjin:
	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -233,7 +251,7 @@ type core_match_extra <python decorator="dataclass(frozen=True)"> = {
   ?fix: string option;
 
   ?dataflow_trace: match_dataflow_trace option;
-  engine_kind: engine_kind;
+  engine_kind: engine_flavor;
```
aryx, line 254: Same comment that in the other PR. the difference between kind and flavor is confusing.
Use just one, and since the field is called engine_kind, I would use engine_kind for the type.

aryx (changes requested):

semgrep_output_v1.atd
```diff
@@ -165,7 +165,25 @@ type error_severity
...
+   
+   OSS and PRO could have clearer names, but for backwards compatibility
+   we're leaving them as is *)
+type engine_flavor
+     <ocaml attr="deriving show">
+     <python decorator="dataclass(frozen=True)"> = [
+  | OSS
+  | PRO
+  | PRO_SPECIFIC
+  | PRO_SPECIFIC_INTRAFILE_TAINT
```
IagoAbal, line 183: Could this be a separate data type? `PRO_SPECIFIC of pro_specific_kind` ? I'm not great at naming but `PRO_ONLY` is shorter.

IagoAbal: Perhaps it could even be `OSS | PRO of pro_kind` ?

semgrep_output_v1.atd
emjin: I would like to separate engine_type in general (for rules, for checking the requested engine, etc) from engine_type for findings (which attaches information about which feature was used). Hopefully this rename is cleaner. I was hesitate to rename `engine_kind` in the atd because I'm worried about backwards compatibility and I think this is reasonable. I did rename it in the OCaml code.

semgrep_output_v1.atd
IagoAbal: But I don't think `PRO_SPECIFIC of something` breaks backwards compatibility, and having a flat structure here doesn't feel right. How many constructors we'll end up having if we later want to track more features? What if a finding involves multiple features and we want to record them all?

aryx (approved):

semgrep_output_v1.atd
```diff
@@ -165,11 +165,38 @@ type error_severity
...
-(* TODO? PRO can actually have "flavors" (see Engine_type.ml) *)
+(* Used for a best-effort report to users about what findings they get with
+   the pro engine that they couldn't with the oss engine.
+
+   Intrafile_taint = requires intrafile taint
+   Interfile_taint = requires interfile taint
+   Other_pro_feature = requires some non-taint pro feature *)
+type pro_feature
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> = [
```
IagoAbal, line 176: Should this be a record with Booleans? May it be easier to extend in the future ? Not a strong opinion, but I'm thinking what happens when we want to track features A and B and a finding can involve both.

semgrep_output_v1.atd
```diff
@@ -165,11 +165,38 @@ type error_severity
...
+  | Intrafile_taint
```
IagoAbal, line 177: `Interproc` would be more accurate here

semgrep_output_v1.atd
```diff
@@ -165,11 +165,38 @@ type error_severity
...
+   PRO = ran with PRO, but we didn't infer that OSS couldn't have found this finding
+   PRO_SPECIFIC = ran with PRO and requires a PRO feature (see pro_feature_used)
+   
+   Note: OSS and PRO could have clearer names, but for backwards compatibility
+   we're leaving them as is *)
+type engine_of_finding
+     <ocaml attr="deriving show">
+     <python decorator="dataclass(frozen=True)"> = [
+  | OSS 
+  | PRO
```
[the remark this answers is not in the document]
emjin: Added a comment

semgrep_output_v1.atd
emjin: I think you're right. I don't love the idea of this being a permanent system for reporting a matrix of pro features used to customers, but it doesn't hurt to be flexible.

semgrep_output_v1.atd
```diff
@@ -165,11 +165,44 @@ type error_severity
   | Info <json name="info">
 ]
 
-(* TODO? PRO can actually have "flavors" (see Engine_type.ml) *)
+(* Used for a best-effort report to users about what findings they get with
+   the pro engine that they couldn't with the oss engine.
+
+   Intrafile_taint = requires intrafile taint
```
aryx, line 171: need to update the comment.

semgrep_output_v1.atd
```diff
@@ -165,11 +165,44 @@ type error_severity
...
+}
+
+(* Report the engine used to detect each finding. Additionally, if we are able to
+   infer that the finding could only be detected using the pro engine, report that
+   the pro engine is required and include basic information about which feature is
+   required.
+
+   OSS = ran with OSS
+   PRO = ran with PRO, but we didn't infer that OSS couldn't have found this finding
+   (* Semgrep 1.64.0 or later *)
```
aryx, line 189: move this comment abouce since semgrep 1.64.0 just before the PRO_REQUORED consturctor in the code.

PR #229, opened 2024-03-05
Add injectable "actions" in scan_config

aryx:
This will enable to modify dynamically old versions of the CLI
(well at least the versions starting from 1.64.0 which is
when this feature was added).
We need to add this ASAP because the more we wait, the less
control we have of old versions of the CLI.

test plan:
make

	  by Semgrep 1.17.0.


aryx:
@nmote this could actually be a good example of use for your pysemgrep RPC; we could implement the action evaluator
in osemgrep and pysemgrep could just fallback to osemgrep RPC for that.

zyannes (approved):

aryx:
@zyannes there will be a good amount of work in the backend to make use of this feature, but at least we can merge this ASAP so we can control versions starting from 1.64.0

PR #230, opened 2024-03-07
feat(secrets): add historical info to findings

salolivares:
This updates the finding type to include historical info. This allows the app to consume the historical scan info.

	  by Semgrep 1.17.0.


kopecs (commented):
lgtm

kopecs (approved):

PR #231, opened 2024-03-15
Add dependency ref to FoundDependency

jarrydlee:
Added an optional field to capture the ref of a dependency if applicable. This allows support for a package refs if a package is pulled in from a git based source.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -917,6 +917,7 @@ type found_dependency = {
   ?line_number: int option;
   (* store child dependency information for calculating path to transitivity*)
   ?children: dependency_child list option;
+  ?ref: string option
```
aryx, line 920: please add a comment before like (* since 1.66.0 *).

aryx: also I have no idea what a ref means, so maybe an example would be useful.

gautambhat (approved):

semgrep_output_v1.atd
jarrydlee: updated naming, and added doc

PR #232, opened 2024-03-18
Add types for Python -> OCaml RPC calls

nmote:
	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1506,3 +1506,39 @@ type ci_env = (string * string) list
...
+
+type apply_fixes_params
+    <python decorator="dataclass(frozen=True)"> = {
+  dryrun: bool;
+  edits: edit list;
+}
+
+type apply_fixes_return
+    <python decorator="dataclass(frozen=True)"> =
+  (int * (int * string list) list)
```
aryx, line 1533: a comment could help to explain what are those int a string

aryx (approved):

semgrep_output_v1.atd
nmote: https://github.com/semgrep/semgrep-interfaces/pull/234

PR #234, opened 2024-03-19
Make apply_fixes_return a record

nmote:
Requested in #232

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1528,6 +1528,11 @@ type apply_fixes_params
   edits: edit list;
 }
 
+(* The first int is the number of modified files.
```
aryx, line 1531: I think a record would be better than this complex tuple. Then you could use
nice field names in the OCaml and Python code.

aryx (approved):

aryx (approved):

PR #237, opened 2024-03-22
feat(secrets): add run custom validators as engine param

salolivares:
This PR adds the `run_custom_validators` engine param in order to inform the CLI whether or not to run custom validator rules.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1260,6 +1260,8 @@ type engine_configuration = {
     ~generic_slow_rollout: bool;
     (* from 1.63.0 *)
     ?historical_config: historical_configuration option;
+    (* from 1.66.0, used to decide whether we want to run custom validator rules *)
+    ?run_custom_validators: bool option;
```
salolivares, line 1264: @kopecs i've blindly added `run_custom_validators` to `engine_configuration`. Two things:
* Don't feel to strongly about the name 
* Could see an argument for adding this to `scan_configuration` instead. 

^ any strong opinions on this?

kopecs: I think the location is fine. I am a bit wary of making this just a boolean flag. Not sure if we maybe want to give people the option from say, { none, only pro rules, pro + our custom, <something more permissive?> }.

aryx: we probably want to add this field in the `features` record instead of here (see the inherit features a few lines above).

aryx (changes requested):

salolivares:
no longer needed.

PR #238, opened 2024-03-27
Associate .phtml extension to PHP

p4p3r:
Sometimes PHP application use the .phtml extension, especially for files which contain a mix of html and php code instead of just php.

In the PHP specs, everything that is not within the `<?php` and  `?>` tags is considered unstructured text:
```
script:
   script-section
   script script-section

script-section:
   text start-tag statement-list end-tag text

start-tag:
   <?php
   <?=

end-tag:
   ?>

text:
   arbitrary text not containing any of start-tag sequences
``` 
Our grammar seems to follow this rather closely.

Associating .phtml with PHP basically means:
- we will be able to apply PHP rules to .phtml files, without parsing errors (thanks to the arbitrary text syntactic element)
- we will not parse the arbitrary text as html (which it usually is), we'd need something like extract mode for that, but it's probably a minor issue with smaller impact than having to use generic mode for every .phtml files like today

I did some *cursory* testing on .phtml files to confirm that they parse fine (using pointer from [](https://github.com/semgrep/semgrep/issues/10009)) and it seems to be the case.
Here's an example of rule for PHP running on a phtml file: https://semgrep.dev/playground/s/d8EkP

	  by Semgrep 1.17.0.


p4p3r:
@aryx @mjambon, do I need to update the Semgrep repo too?

aryx (approved):

aryx:
Yes you need a PR in semgrep that update the cli/src/semgrep/semgrep_interface submodule to point to the latest version,
otherwise we will compile semgrep with the older version of semgrep_interface (which contains the old version of the generated Language.ml, which is linked in semgrep).

PR #239, opened 2024-03-27
feat(secrets): new origin policy for not community

kopecs:
This is needed since we are updating the way the origin based filtering for validation works, and changing it from a whitelist of [pro_rules] to a blacklist of [community], i.e., effectively, [pro_rules, custom, None].

	  by Semgrep 1.17.0.


salolivares (approved):
LGTM!

salolivares:
@kopecs I'd expect more generated files in this PR.

kopecs:
I think it is just because it is `semgrep_metrics.atd` instead of one of the ones we need more languages for.

PR #240, opened 2024-03-28
chore(pro-indicators): send engine_kind to the app

emjin:
Send the engine kind to the app so that we can use it to display when findings require the pro engine.

	  by Semgrep 1.17.0.


amchiclet (approved):

semgrep_output_v1.atd
```diff
@@ -1310,6 +1310,8 @@ type finding = {
   ?validation_state: validation_state option;
   (* Added in semgrep 1.65.0 see comments in cli_match_extra *)
   ?historical_info: historical_info option;
+  (* Added in semgrep 1.70.0 *)
```
aryx, line 1313: you mean 1.69.0? We're still at 1.68.0

PR #241, opened 2024-04-01
feat: add codeql

brandonspark:
	  by Semgrep 1.17.0.

### DIFF
--- a/Language.ml
+++ b/Language.ml
@@ -30,6 +30,7 @@ type t =
 | Python2
 | Python3
 | Python
+| Ql
 | R
 | Ruby
 | Rust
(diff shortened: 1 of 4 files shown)


tpetr (approved):
- why ql vs codeql? is one the language and the other the brand name?
- curious what's the difference between `.ql` and `.qll` extensions?

overall lgtm

brandonspark:
> * why ql vs codeql? is one the language and the other the brand name?
yep, that's correct

> * curious what's the difference between `.ql` and `.qll` extensions?
`.ql` referring to a query file, and `.qll` a library file to assist a query

PR #242, opened 2024-04-01
Add hex ecosystem for SCA

jarrydlee:
`Mix` was incorrectly used as the ecosystem, even though it is a build tool for elixir, not the ecosystem. We're adding `hex` so that we can correctly associate findings with the correct package manager.

	  by Semgrep 1.17.0.


mmcqd (approved):

semgrep_output_v1.atd
```diff
@@ -885,7 +885,9 @@ type ecosystem <python decorator="dataclass(frozen=True)"> <ocaml attr="deriving
   | Nuget <json name="nuget">
   | Pub <json name="pub">
   | SwiftPM <json name="swiftpm">
+  (* Deprecated: Mix is a build system, should use Hex, which is the ecosystem *)
```
aryx, line 888: Can we juste delete Mix instead of adding the deprecated notice?
Is there any rules/code relying on Mix?

PR #243, opened 2024-04-12
feat(osemgrep): add structs and metrics to prepare osemgrep sarif format

amchiclet:
I'm trying to have osemgrep handle `semgrep scan --sarif`, so in the end, the new flow will look like this.

`pysemgrep -> semgrep-core -> pysemgrep post-processes results -> RPC to osemgrep to format results`

To support this RPC, I added a new struct that takes in pretty much what ocaml sarif takes in as input.

To support a safe and incremental migration, I also added metrics to keep track of the performance and correctness.

If you'd like to see concretely how these new metrics and data structs are used, they are in this WIP branch.
* The RPC call [would look like this.](https://github.com/semgrep/semgrep/blob/82016ba507594fe33cde1aad9b30e88a54cde1ce/cli/src/semgrep/formatter/osemgrep_sarif.py#L77-L83)
* The metrics populating code [would look like this.](https://github.com/semgrep/semgrep/blob/82016ba507594fe33cde1aad9b30e88a54cde1ce/cli/src/semgrep/formatter/osemgrep_sarif.py#L145-L154)

On performance:

You may notice that the rules are passed as a path, which means I'm writing rules to a file in python, then osemgrep reads that file. This is the bottle neck for this particular RPC.

I tested this with `semgrep scan --config r/all` (which has ~3000 rules) on the semgrep repo , and the RPC took almost 2 seconds from ~90 seconds of matching time, so I think the performance hit is acceptable.

For a smaller example, I tested `semgrep scan --config p/python` (which has ~300 rules) on the `cli` directory, and the RPC took less than 0.5 seconds from 5 seconds of matching time. Again, I think the performance hit is acceptable.

I think it's simple and the trade off is worth it, but if you have thoughts about easy performance wins, I'm happy to experiment with any of your ideas a bit.

PR Checklist:
	  by Semgrep 1.17.0.


semgrep_metrics.atd
```diff
@@ -57,6 +57,12 @@ type payload = {
      *)
     ~parse_rate <ocaml mutable>: (string (* lang *) * parse_stat) list
                                  <json repr="object">;
+
+    (* Metrics related to osemgrep migration. It's intentionally made optional
+     * so that it's easier to remove the field without breaking backward compatibility
+     * once the migration is complete.
+     *)
+    ?osemgrep <ocaml mutable>: osemgrep_metrics option;
 }
```
mjambon, line 66: This change is fine but I don't know why all these fields are marked as mutable. It's rarely necessary.

mjambon (approved):

semgrep_metrics.atd
mjambon:
```
commit 2b0b774ee493045b83b6ec6447462e514bc0d83c
Author: Calascibetta Romain <romain.calascibetta@gmail.com>
Date:   Sun Jun 11 12:24:29 2023 +0200

    Set metrics's fields to mutable fields
```
Unfortunately, there's no explanation for this. Please add a note such as `(* TODO: explain why the fields are marked as mutable since this is unusual. *)`

amchiclet: I did not know that. I'll try testing them without the mutable fields, and if that works, then I'll merge.

amchiclet: And I'll add the comments.

amchiclet:
Done.

mjambon (approved):

PR #244, opened 2024-04-15
Add comment about the use of mutable in semgrep_metrics.atd

aryx:
test plan:
make

make in semgrep

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -35,8 +35,11 @@ type lang = string
 (* Entry point *)
 (*****************************************************************************)
 
-(* TODO: explain why the fields are marked as mutable since this is unusual. *)
-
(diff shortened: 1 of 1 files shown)


amchiclet (approved):
The comments make sense, but I don't understand why the generated code doesn't change. I would expect something like the example in the documents: https://atd.readthedocs.io/en/latest/atdgen.html#field-mutable

aryx:
because we generated code for semgrep_metrics.atd is not part of the repo; only the code for semgrep_output_v1.atd is, for convenience so that other repos like semgrep-app can just use the submodule and have access to the python and TS binding for semgrep_output_v1.atd. Right now only semgrep is using semgrep_metrics.atd so we don't need the generated code in the submodule.

PR #245, opened 2024-04-17
Add TODO comments in semgrep_metrics.atd

aryx:
test plan:
make

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -9,9 +9,19 @@
  * translated originally from semgrep/cli/.../metrics.py
  *
  * TODO:
- * - some optional arguments should be removed when `osemgrep` replace
- *   `semgrep`. Currently, we tried to fit on tests and don't update the
- *   JSON output. We tried to fit on the Python side too.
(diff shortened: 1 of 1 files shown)


semgrep_metrics.atd
```diff
@@ -9,9 +9,19 @@
...
  *
  * TODO:
- * - some optional arguments should be removed when `osemgrep` replace
- *   `semgrep`. Currently, we tried to fit on tests and don't update the
- *   JSON output. We tried to fit on the Python side too.
+ * - some optional arguments could be removed when `osemgrep` replace
+ *   `pysemgrep`. Currently, we try to fit the tests and don't update the
+ *   JSON output.
+ * - specify the schema for the answer from https://metrics.semgrep.dev
+ *   Here is an example of answer:
```
aryx, line 16: @amchiclet I can reproduce those errors with
running `./bin/osemgrep --experimental --config semgrep.jsonnet --error --strict --exclude tests --debug` in the semgrep repo (essentially what `make check` is running).
Looks like osemgrep can't even send metrics currently ...

PR #246, opened 2024-04-18
Adding a new metric for the number of scanned diff targets in inter-file diff scan mode

ihji:
	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -179,6 +179,11 @@ type pro_features = {
     (* The value is determined by the `--diff-depth` option, and it's utilized
        for inter-file differential scanning. Since semgrep 1.46 *)
     ?diffDepth <ocaml mutable>: int option;
+    (* The number of scanned files per language for inter-file diff scan mode.
+       This number represents the count of changed files and their limited
+       dependencies. Since semgrep 1.70 *)
+    ?numInterfileDiffScanned <ocaml mutable>: (string (* lang *) * int) list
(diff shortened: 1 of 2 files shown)


semgrep_metrics.atd
```diff
@@ -179,6 +179,11 @@ type pro_features = {
     (* The value is determined by the `--diff-depth` option, and it's utilized
        for inter-file differential scanning. Since semgrep 1.46 *)
     ?diffDepth <ocaml mutable>: int option;
+    (* The number of scanned files per language for inter-file diff scan mode.
+       This number represents the count of changed files and their limited
+       dependencies. Since semgrep 1.70 *)
+    ?numInterfileDiffScanned <ocaml mutable>: (string (* lang *) * int) list
+                                              <json repr="object"> option;
```
IagoAbal, line 186: Why not a new field in `ci_scan_complete_stats` ?

aryx: Yes there is a design decision each time of whether put the info in metrics or in the CLI output ... Not sure what is the rule. Maybe we should document it somewhere, like at the top of this pro_features data strutcture.

aryx (approved):

semgrep_metrics.atd
ihji: We already have an output field for the list of [scanned](https://github.com/semgrep/semgrep-interfaces/blob/main/semgrep_output_v1.atd#L617) files. The metric we've added here can be easily generated from that list.

PR #247, opened 2024-04-26
Add Critical/High/Medium/Low for rule and match severity

aryx:
test plan:
make
see related PR in semgrep

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -145,6 +145,15 @@ type match_severity
   | Info <json name="INFO">
   | Experiment <json name="EXPERIMENT">
   | Inventory <json name="INVENTORY">
+  (* since 1.71.0, meant to replace the cases above where
+   * Error -> High, Warning -> Medium, and Info -> Low,
+   * and Experiment and Inventory should be removed.
```
chmccreery, line 150: Is there more information on the removal of Experiment and Inventory? I mapped these to "low" on the app side (marked findings that had severity "experiment" as "low" severity.

semgrep_output_v1.atd
```diff
@@ -1282,7 +1291,11 @@ type finding = {
   end_column: int;
 
   message: string;
-  (* int|string until minimum version exceeds 1.32.0, then string *)
+  (* int|string until minimum version exceeds 1.32.0, then int?
```
chmccreery, line 1294: Does minimum version exceed 1.32.0 yet? How far away from that are we?

semgrep_output_v1.atd
```diff
@@ -1282,7 +1291,11 @@ type finding = {
...
+   * Error = 2, Warning = 1, Experiment = 4, otherwise 0
```
chmccreery, line 1295: Above it says deprecating Experiment but here it looks like it still has its own int?

semgrep_output_v1.atd
aryx: I think Inventory was used for CAI (Code Asset Inventory). I don't think we used it anymore. Maybe @p4p3r knows.
Experiment, I'm not sure what it is. Again maybe @p4p3r or @minusworld remember.

semgrep_output_v1.atd
aryx: Not sure what it is ... Jonas was managing that and I don't know who took over this.

semgrep_output_v1.atd
aryx: I'm just commenting here what pysemgrep is currently doing. I'll be happy to remove that.
But really I would be happier if we were switching from an int to a string (actually an enum, that is the one in match_severity) actually going forward.

semgrep_output_v1.atd
p4p3r: Code/SR doesn't use these values at the moment but it would be useful to keep one of them (experiment?) because we'd like to revive the experimentation framework at some point.

chmccreery: I think that SSC maybe still uses experiment based on looking at app code. But if not, this would be a great opportunity IMO to get rid of both of these fields and if we want to spin up an experimentation framework down the line do it without overloading "severity". cc @misha-semgrep for insight maybe on the SSC Experiment severity

semgrep_output_v1.atd
chmccreery: Ok I found it in the code. It's 1.31.0 and was bumped to that 3 months ago. It feels like it should be safe to bump this above 1.32.0 in that case? I don't think anyone "owns" this right now sadly but I can dig a bit.

p4p3r (dismissed):
lgtm

p4p3r (approved):
lgtm

PR #248, opened 2024-04-26
feat(secrets): add `product_ignored_files` to engine configuration

salolivares:
Scope document: https://www.notion.so/semgrep/Per-product-targeting-64e5029758334dbbaa8a930b6bf3aefe?pvs=4

This PR adds a new optional field `product_ignored_files` to `engine_configuration`. This will kick off work to send product specific ignore lists to the CLI.

For example, here's how the generated typescript type looks:

```ts
export type Glob = string

export type ProductIgnoredFiles = Map<Product, Glob[]>

export type EngineConfiguration = {
  autofix: boolean;
  deepsemgrep: boolean;
  dependency_query: boolean;
  ignored_files: string[];
  product_ignored_files?: ProductIgnoredFiles;
  generic_slow_rollout: boolean;
  historical_config?: HistoricalConfiguration;
}
```

---

	  by Semgrep 1.17.0.


kopecs (approved):

kopecs (commented):
Actually, this probably shouldn't be in `engine_configuration`.

salolivares:
> Actually, this probably shouldn't be in `engine_configuration`.

What would you suggest? Given the scan response type:

```
type scan_response = {
    info: scan_info;
    config: scan_configuration;
    engine_params: engine_configuration;
}
```

would it make more sense scan_configuration or somewhere entirely different?

salolivares:
@kopecs i'll move it into scan config for now, and await @aryx's opinion before merging.

salolivares:
> I'm not sure why we have both scan_configuration and scan_config.

`scan_config` isn't used yet. Guessing we're going to move to it in the future. For now, i'll drop in `scan_configuration`.

aryx:
Can we use the product enum like you proposed at the beginning?
You can use (product * glob list) list; instead?
(where glob is just an alias for string, with type glob = string; 

You can use (product * glob list) list; it's just that you must not add after the json repr=object after.
cc @mjambon for this ATD restriction.

semgrep_output_v1.atd
```diff
@@ -1240,10 +1240,18 @@ type scan_info = {
...
+  ~sast: string list;
+  ~sca: string list;
+  ~secrets: string list;
+}
+
 (* config specific to the scan, eg *)
 type scan_configuration = {
     rules: raw_json; (* can we type this better *)
     inherit triage_ignored;
+    (* from 1.71.0 *)
```
aryx, line 1253: maybe better in engine_configuration next to ~ignored_files?

I'm also a bit confused why we have scan_config and scan_configuration ... this was part of some refactoring that Jonas started I think but we didn't finish it.
I'm not sure what goes in scan_configuration vs engine_configuration ... they are all useful to configure the engine.

aryx (changes requested):

semgrep_output_v1.atd
salolivares: addressed in [58f73b3](https://github.com/semgrep/semgrep-interfaces/pull/248/commits/58f73b39e0d1a544892581f73a64a85e43ebaacf) and [8e11f3c](https://github.com/semgrep/semgrep-interfaces/pull/248/commits/8e11f3cc0bf43c5f8a0205c0a591627fb5e0bfa4)! 

I dropped `<json repr="object">` and that worked!

semgrep_output_v1.atd
```diff
@@ -1254,10 +1256,19 @@ type historical_configuration = {
     ?lookback_days: int option;
 }
 
+type glob = string
+
+type product_ignored_files = (product * glob list) list
```
aryx, line 1261: @mjambon what are the consequences if tomorrow we add a new product, so the product enum gets a new entry.
What will do old versions of semgrep if suddenly they get in those lists a new product? The element is skipped?
An error is generated? What about those association lists?

aryx (approved):

semgrep_output_v1.atd
mjambon: We could implement an error-tolerance mechanism similar to what tree-sitter does: parsing errors bubble up until reaching a node that can be skipped safely i.e. a list element, optional, or nullable. In the case of an invalid `product` in `(product * glob list) list`, the invalid JSON pair `["New_product_kind", [...]]` would be removed from the list and kept aside as an error. This work would have to be done for each programming language that wants this feature (atdpy for Python, atdgen for OCaml) except that I don't want to even try to add such features to atdgen because it's too complicated. It would be reasonable to do in atdpy, atdts, or a [future atdml reimplementation of atdgen](https://github.com/ahrefs/atd/issues/313) that simplifies things.

aryx: Will atdcheck report those regressions? I mean if I add a new enum in Product, will atdcheck complains that suddenly the product_ignored_files becomes backward incompatible?

aryx: So in the end, using strings is actually more robust to evolution atd-wise; but anyway it would just delay the issue
because we would need to convert those strings in actual Product type later

PR #249, opened 2024-04-30
feat(osemgrep): add dataflow-traces as a parameter to SARIF RPC

amchiclet:
Helps SAF-1020

Osemgrep's implementation of SARIF needs to acknowledge --dataflow-traces. This PR adds that field to the RPC, so it can be passed to the osemgrep formatter.


amchiclet:
	  by Semgrep 1.17.0.

aryx (approved):

aryx (approved):

aryx:
I don't know if there's a way to disable the backward compatibility check for parts of the spec.
In theory for changes in semgrep-core JSON output we also don't care for backward compabitlity.

amchiclet:
Yeah, sadly I don't know either whether we can bypass a required check for a specific commit or not. :(

PR #250, opened 2024-05-03
fix(secrets): change product to string

salolivares:
Changing the ATD definition from
```
type product_ignored_files = (product * glob list) list
  (* We omit the usual <json repr="object"> else we get a "keys must be strings" error *)
  <python repr="dict">
  <ts repr="map">
```
to
```
type product_ignored_files = (string * glob list) list
  <json repr="object"> 
  <python repr="dict">
  <ts repr="map">
```

Omitting <json repr="object"> makes the product ignores json into this:
```
    "product_ignored_files": [
      [
        "sast",
        []
      ],
      [
        "secrets",
        []
      ],
      [
        "sca",
        []
      ]
    ],
```

This makes it very developer UN-friendly when dealing with data structure in `semgrep-app`.

---

	  by Semgrep 1.17.0.


salolivares:
> Checking backward compatibility of semgrep_output_v1.atd against past version v1.71.0
> --- before.txt
> +++ after.txt
> @@ -0,0 +1,7 @@
> +[0dceee86] Incompatibility in both directions:
> +Incompatible kinds of types: list/array is now a map.
> +The following types are affected:
> +  engine_configuration
> +  product_ignored_files
> +  scan_response
> +
> ERROR: semgrep_output_v1.atd is not backward compatible with v1.71.0

I think this is okay since this field is brand new and we don't use it yet.

aryx (changes requested):
No.

I don't understand the complaint. Having product here is better because it's a product, it's not any string, it must be a product.

We want types.

I don't understand the complain about UN-friendlyness in semgrep-app ... semgrep-app can use the generated code to access those data.

aryx:
Adding @mjambon who might have an opinion.

salolivares:
@aryx figured i'd get push back. Appreciate your thoughts! I'll forgo these changes then. 

I was torn between type safety and developer experience on the app side. Really wish json option was supported.

aryx:
On the app side, they should start to use the generated code here to manipulate this "product" type. Then really it does
not matter how this product type is incoded in json.

PR #251, opened 2024-05-07
Update rule schema to allow AWS secrets validator

mmcqd:
See https://github.com/semgrep/semgrep-proprietary/pull/1483
	  by Semgrep 1.17.0.


kopecs (approved):
Could you also remove the `auth` field from http-request-content? Otherwise looks good, and we can add additional fields if we want more service/endpoint support for some reason.

aryx (commented):
Please modify also rule_schema_v2.atd!

PR #253, opened 2024-05-22
chore: Add can_batch_upload feature flag

zyannes:
	  by Semgrep 1.17.0.

We want to be able to support batch uploading in the CLI/backend -- the CLI can arbitrarily chunk the findings from a scan and upload in batches. In order to do this, we want to:
- Add a `batch_info` field to the ci_scan_results to indicate the progress of uploads and tell the backend how many upload batches to expect
- Add a `can_batch_upload` feature flag so the backend can inform the CLI whether to use batch uploading or not


semgrep_output_v1.atd
```diff
@@ -1030,6 +1030,7 @@ type features = {
    ~autofix: bool;
    ~deepsemgrep: bool;
    ~dependency_query: bool;
+   ~can_batch_upload: bool;
```
aryx, line 1033: please add a (* since 1.74.0 *) comment

aryx (approved):

semgrep_output_v1.atd
zyannes: Added!

semgrep_output_v1.atd
```diff
@@ -1362,6 +1364,11 @@ type finding_hashes = {
 (* CI scan results *)
 (* ----------------------------- *)
 
+type batch_info = {
```
aryx, line 1367: I talked with @akuhlens a little bit last week about this project, and it feels a bit sad to have
to support use cases where companies want semgrep to upload 10 000 findings ... what's they gonna do with
those 10000 findings? It's not a good use case; Also I don't really understand either why we need to split those findings results in batches ... it feels like we reinventing packet management and TCP. 
But, whatever.

aryx (approved):
I would change
can_batch_upload to use_batch_upload, to convey
the idea that it's the App telling the CLI to use certain features
(like use autofix, use deepsemgrep).

semgrep_output_v1.atd
zyannes: Yeah I agree, it's not ideal for users to be uploading 10k+ findings. But I do think that supporting batching is good in the long run since it will give the CLI and the backend more flexibility without requiring coordinated releases. 

For example, we may want the CLI to run each product in parallel and upload the results per-product. Similarly, we may want to update the prioritization of result processing, ingesting dependency results as soon as they're uploaded.

aryx: I'm fine with your last 2 use cases, but I still don't see why we need batch_info for that ...
In fact, if what we want is streaming results, then you don't know in advance the num_batches; you'll discover this as we go.

Not a fan of reimplementing our own packet management on top of an existing packet management (TCP), which handles it far better than us. Talking to @akuhlens it seems one motivation is that we timeout sometimes because the message is too big but in that case we can just modulate the timeout in function of the size of the #findings. Also even if 10k is ridiculously high number of findings (I don't see how anybody can make use of that), it's still a small number; it's not 1 trillion finding. We should not timeout sending 10k findings ... people send huge video through the wire and TCP handles it fine. If 10k finding result in huge messages, then maybe we need to work that out. If processing 10k finding in the backend takes too long, then maybe we need to address that instead. I don't think reimplementing TCP is the solution to the problem you're trying to solve.

aryx (changes requested):
I think we need a document explaining more clearly the need for this batch system.

aryx:
Happy to have a meeting about this; I think we should brainstorm alternative solutions before going the route of packet management.

DrewDennison:
I did API design session with @zyannes and we're looked at how AWS does this for large uploads: 
https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

Principle design decisions
1. We don't want to rely on knowing now many batches there are
2. We want to make sure this is future proof so batch size should not be fixed and should gracefully migrate to streaming later (i.e. batch size of 1)

Proposal:
Delete `current_batch` and `num_batches`
Add `batch_type` as an ENUM of `partial|last`

System then works this way:
CLI POST to `/start` with any job metadata available at start of run
CLI POST one or more uploads to `/upload`. Last post MUST set `batch_type` to `last`
CLI POST to `/complete` with any job metadata only available at end of run e.g. num findings, run time

zyannes:
@DrewDennison I'd suggest we keep `current_batch` but rename it to `page_num`. It would be good to have the uploads include a unique identifier so that the backend can differentiate between unique attempts and remain idempotent. Consider the following example:

1. CLI POST to `/results` with `page_num=1, batch_type=partial` -> receives 201
2. CLI POST to `/results` with `page_num=2, batch_type=last` -> receives 500 (due to unexpected server error)
  a. At this point we don't know whether the batch was ingested successfully. The CLI will retry on 5xx.
3. CLI **retries** POST to `/results` with `page_num=2, batch_type=last` -> receives 201

We could do without it (and use checksum or server-side counter), but it would be beneficial to have the CLI intentional about which batch it is uploading.

zyannes:
Whoops just saw you're comment was updated and includes `part_number`. Disregard my comment

semgrep_output_v1.atd
```diff
@@ -1051,6 +1051,8 @@ type features = {
    ~autofix: bool;
    ~deepsemgrep: bool;
    ~dependency_query: bool;
+   (* since semgrep 1.74.0 *)
+   ~use_batch_upload: bool;
```
akuhlens, line 1055: Does this need to be optional for backwards for compatibility reasons? Just double checking, I think the answer is no.

aryx: Yes any change to semgrep_output_v1.atd but support backward compatibility; one of the CI check above actually enforces that. The use of ~xxx support backward compability because if the field is not present, it is assumed
to have a default value (I think for boolean the default is false).

zyannes: Good catch. It doesn't strictly need to be optional, but it feels like a good practice. Updated in latest commit

DrewDennison (approved):
I like this new API design better for the reasons listed in my PR. I acknowledge there is a larger discussion around why this is needed and also what the path to a simpler upload looks. I'll send a follow up sync invite

but this code looks right and backwards compatible to me

semgrep_output_v1.atd
```diff
@@ -1405,6 +1418,9 @@ type ci_scan_results = {
    (* since semgrep 1.38.0 *)
    (* this data was originally sent to /complete, but we want to start sending it /results *)
    ?dependencies: ci_scan_dependencies option;
+
+   (* since semgrep 1.74.0 *)
+   ?batch_info: batch_info option;
```
aryx, line 1423: What is batched? Just the finding list? also the ignores? What about searched_paths? Also split in parts? What about
rule_ids?
And all the other fields, they will be present in each batch?

Seems like it would be better to use a different record for those batch things.

tpetr:
> CLI POST one or more uploads to /upload. Last post MUST set batch_type to last
why this approach vs. having the complete endpoint accept the list of expected part numbers?

aryx:
Not sure what's the status of this, but let's close in doubt. I'm trying to clean the list of opened PRs.

PR #254, opened 2024-05-29
Add comments for CliLoc & CliCall

underyx:
	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -396,37 +394,60 @@ type svalue_value <python decorator="dataclass(frozen=True)"> = {
...
+ *  taint_sink = CliCall("bar()" @l9, ["v"], CliLoc "sink(v) @l5")
+ *)
+
 type match_dataflow_trace <python decorator="dataclass(frozen=True)"> = {
   ?taint_source: match_call_trace option;
-  (* Intermediate variables which are involved in the dataflow. For taint, this
+  (* Intermediate variables which are involved in the dataflow. This
    * explains how the taint flows from the source to the sink. *)
   ?intermediate_vars: match_intermediate_var list option;
-  (* For now, omitting the taint sink, since it's redundant data except for
```
aryx, line 406: @IagoAbal I've removed this comment as it appears to be incorrect.

semgrep_output_v1.atd
```diff
@@ -396,37 +394,60 @@ type svalue_value <python decorator="dataclass(frozen=True)"> = {
...
+  | CliLoc of loc_and_content
+  | CliCall of (loc_and_content * match_intermediate_var list * match_call_trace)
 ] <ocaml repr="classic">
 
 
-(* EXPERIMENTAL *)
+(* This type happens to be mostly the same as a loc_and_content for now, but
+ * it's split out because Iago has plans to extend this with more information
+*)
 type match_intermediate_var <python decorator="dataclass(frozen=True)"> = {
```
IagoAbal, line 446: Also, this right now is guaranteed to refer to a variable/l-value, whereas `loc_and_content` could be an arbitrary thing.

IagoAbal (approved):

semgrep_output_v1.atd
IagoAbal: I'm not sure what the comment meant exactly, but inter-proc is now the default. I guess it meant that in intra-proc, the trace for the `taint_sink` will just be a `CliLoc` and it will be the same loc as the finding, so in that sense it is redundant.

PR #255, opened 2024-05-29
feat: add nonkeyword_attributes_order_matters rule option in rule schema atd

liukatkat:
This PR adds a new rule option `nonkeyword_attributes_order_matters`, which allows users to make decorator/ non-keyword attributes matching stricter, to the rule schema atd file. The default matching for attributes is order-agnostic, but if this rule option is set to `true`, non-keyword attributes (e.g. decorators in Python) will be matched in order, while keyword attributes (e.g. static, inline, etc) are not affected.

	  by Semgrep 1.17.0.

### DIFF
--- a/rule_schema_v2.atd
+++ b/rule_schema_v2.atd
@@ -232,7 +232,7 @@ type fix_regex = {
  *)
 type rule_options = {
   ?constant_propagation: bool option;
-  ?decorators_in_order: bool option;
+  ?nonkeyword_attributes_order_matters: bool option;
   ?symbolic_propagation: bool option;
 
   ?taint_unify_mvars: bool option;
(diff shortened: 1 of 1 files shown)


amchiclet:
Looks good! I've added Pad to review too, since it's something that touches the interface.

@aryx here's where this change is being used: https://github.com/semgrep/semgrep-proprietary/pull/1610

amchiclet (approved):

rule_schema_v2.atd
```diff
@@ -232,6 +232,7 @@ type fix_regex = {
  *)
 type rule_options = {
   ?constant_propagation: bool option;
+  ?nonkeyword_attributes_order_matters: bool option;
```
aryx, line 235: please put it at the end or with another option that has similar semantic.

aryx (changes requested):

aryx (approved):
I don't know if amarin told you, but often reviewers put a few comments and request and approve
to avoid an extra round of review, so that you can just address the comments and merge.

amchiclet:
Looks like we need to rebase. There's a new change on the main branch.

PR #258, opened 2024-06-07
include deployment id

ajbt200128:
We want to start including the deployment id when reporting metrics. This will be especially helpful for extensions, as then we can track # of logged in users by deployment

	  by Semgrep 1.17.0.


aryx (approved):
You'll need to update OSS/Privacy.md too and be vocal in the changelog.

ajbt200128:
will do that in the PR that uses this change

PR #260, opened 2024-06-18
feat: add `when` to rule syntax

liukatkat:
This PR adds `when` to the rule syntax.

	  by Semgrep 1.17.0.

### DIFF
--- a/rule_schema_v2.atd
+++ b/rule_schema_v2.atd
@@ -352,6 +352,8 @@ type formula = {
 
   (* NEW: since 1.74 *)
   ?fix: string option;
+
+  ?when_: string option;
 }
 <json adapter.ocaml="Rule_schema_v2_adapter.Formula">
 


rule_schema_v2.atd
```diff
@@ -352,6 +352,8 @@ type formula = {
 
   (* NEW: since 1.74 *)
   ?fix: string option;
+
+  ?when_: string option;
```
aryx, line 356: you probably need a <json name="when"> like we do for match_ above:
```
     ?match_ <json name="match">: formula option;
```

aryx (commented):
I think there was a debate wether putting when as a formula, or as a condition
inside where.

rule_schema_v2.atd
brandonspark, line 356: would also be nice to add a `(* NEW: since ... *)`!

you can just find the most recent version of Semgrep here https://github.com/semgrep/semgrep and increment that

aryx:
I think @IagoAbal thought of a way to have a v0 where when would be a metavariable-cond but I don't remember the details. @IagoAbal do you remember?

liukatkat:
i placed `when` at `pattern`'s level because i thought putting `when` under `where` is a bit unintuitive. it might also be a bit confusing since all of the fields under `where` are metavariables related but you can use `when` without them. i think it should be an easy change from putting it at the top level to under `where` though.

liukatkat:
i also talked to um earlier and he suggested that ultimately we might be able to combine `where` condition and `when`. say we have a metavariable condition, on top of the existing information we can also use control-flow information to check the condition. so we can abstract away the fact that we use control-flow from the users and also make where more general. though we weren't sure if we should be doing this in v0.

liukatkat:
thanks for looking at the PR btw!!

rule_schema_v2.atd
```diff
@@ -352,6 +352,8 @@ type formula = {
 
   (* NEW: since 1.74 *)
```
IagoAbal, line 353: What about syntax v1?

IagoAbal:
BTW I wonder why we have `strip` and `base` in `comparison`, those are deprecated already in v1 and should not be part of v2 I think.

IagoAbal:
> i placed `when` at `pattern`'s level because i thought putting `when` under `where` is a bit unintuitive. it might also be a bit confusing since all of the fields under `where` are metavariables related but you can use `when` without them. i think it should be an easy change from putting it at the top level to under `where` though.

The `when` as a conditional would be the same as `comparison`, `comparison` is not specific to any metavariable either. Although as a conditional, maybe we wouldn't even need a new operator and we could just use `comparison`.

IagoAbal:
TBH I would first implement some draft in semgrep-core alone and do the interface changes later, they are not needed for early experimentation.

rule_schema_v2.atd
aryx: This file is just about syntax v2 (hence the filename).

IagoAbal:
> I think @IagoAbal thought of a way to have a v0 where when would be a metavariable-cond but I don't remember the details. @IagoAbal do you remember?

My preliminary thoughts about the "when" was that we could just extend "svalue analysis" and `id_svalue` in `id_info` to store path conditions. E.g., given `foo(x); if (x > 1) { foo(x); }`,  the `id_svalue` of the `x` in the `if`'s then-branch could capture `x > 1`. Then matching `foo($X)` plus `comparison: $X > 1` would filter out the `foo(x)` before the `if`. We could even resolve `x > 0` because `x > 1 -> x > 0`.

liukatkat:
> The `when` as a conditional would be the same as `comparison`, `comparison` is not specific to any metavariable either.

i thought `comparison` is specific to metavariables because when i have an example like the one below, it doesn't match when i use a rule without metavariables.

```
def foo2():
   x = 0
   foo(x)
```

this rule doesn't match `foo(x)`:
```
rules:
- id: find-bad-stuff
  severity: ERROR
  languages: [python]
  message: |
    Don't put bad stuff!
  match:
    pattern: foo(x)
    where:
      - comparison: x == 0
```

but this rule does:
```
rules:
- id: find-bad-stuff
  severity: ERROR
  languages: [python]
  message: |
    Don't put bad stuff!
  match:
    pattern: foo($X)
    where:
      - comparison: $X == 0
```

liukatkat:
> > I think @IagoAbal thought of a way to have a v0 where when would be a metavariable-cond but I don't remember the details. @IagoAbal do you remember?
> 
> My preliminary thoughts about the "when" was that we could just extend "svalue analysis" and `id_svalue` in `id_info` to store path conditions. E.g., given `foo(x); if (x > 1) { foo(x); }`, the `id_svalue` of the `x` in the `if`'s then-branch could capture `x > 1`. Then matching `foo($X)` plus `comparison: $X > 1` would filter out the `foo(x)` before the `if`. We could even resolve `x > 0` because `x > 1 -> x > 0`.

I see! During some discussions a few weeks ago about `pattern-when`, we also talked about matching statements that do not contain any variables, e.g. if we have `if (x == 0) { print("hi") } else { print("hi") }` and a rule that tries to match `print("hi")` `when: x == 0`, we should still be able to match the first `print("hi")`. so we thought maybe we cannot achieve that by annotating `id_svalue`. do we care about those cases in v0?

brandonspark:
> i thought comparison is specific to metavariables because when i have an example like the one below, it doesn't match when i use a rule without metavariables.
Indeed, this example should not match, because there is no binding to `x` -- it's not understood by Semgrep. This doesn't mean that `comparison` must always have metavariables, though. Consider the following Playground example: https://semgrep.dev/playground/s/6JZ2w

With respect to @IagoAbal's suggestion -- this came up early in the design process, but part of the work was scoped to include matching  _arbitrary_ statements and CFG nodes. I am still unsure whether this is in scope, being unfamiliar with all of the possible use cases. It may help for us to have another meeting talking about which functionality we think should be aimed for here.

amchiclet:
@IagoAbal @brandonspark @aryx 

I thought the first small change was to support the rule syntax, so everyone has some "skeleton code" to work on and try to experiment with their own ideas.

Given that it seems quite hard to get consensus on an interface change (even though reverting should not be a problem), maybe providing people skeleton code is wasting more of everyone's time and constraining ideas than we all thought. A good exercise for navigating a pool of opinions regardless. :)

I'll try schedule another meeting for all of us.

At this point, I believe Katrina is about to have two draft implementations ready and share with the team what each of them would look like. That will be hopefully be included in the discussion too.

IagoAbal:
> I thought the first small change was to support the rule syntax, so everyone has some "skeleton code" to work on and try to experiment with their own ideas.

Yeah, but **my** understanding was that we just need the support in the Parse_rule* modules, updating the public interfaces can probably wait.

PR #265, opened 2024-06-28
chore: report unparsable rules

brandonspark:
Just makes it so that we report on unparsable rules.

This will be especially useful for using `semgrep test` on generated rules, which may not necessarily parse properly.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -856,6 +856,8 @@ type tests_result = {
    fixtest_results: (string (* target file *) * fixtest_result) list <json repr="object">;
    config_missing_tests: fpath list;
    config_missing_fixtests: fpath list;
+   (* NEW: since v1.79 *)
+   ?config_unparsable: fpath list option;
```
aryx, line 860: I think we should use the config_with_errors below, and define an error type.

aryx (changes requested):

semgrep_output_v1.atd
brandonspark: i was going with the flow since that seemed to be left as `todo`, i couldn't find a lot of comments explaining what your vision for `config_with_errors` was supposed to be

brandonspark: like, are we deprecating `config_missing_tests` and `config_missing_fixtests`?

aryx: I was porting pysemgrep code, so I'm not even fully sure what config_with_errors here and what is the specific format ...
we can't remove config_missing_xxx because many external users are relying on those fields.

aryx (approved):

brandonspark:
@aryx can we merge this even though it's not backwards-compatible? (replaced the `todo` with a real type)

aryx (approved):

PR #266, opened 2024-06-28
add autofix and ignore fields

ajbt200128:
Record how many autofix and ignores have been triggered via IDE since users have asked for this

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -253,7 +253,7 @@ type extension = {
     ?version <ocaml mutable>: string option;
     (* type of extension, i.e. vscode, or intellij *)
     ?ty <ocaml mutable>: string option;
-    (* number of autofixes via code actions *)
+    (* number of autofixes accepted via code actions *)
     ?autofixCount <ocaml mutable>: int option;
     (* how many findings have been ignored *)
     ?ignoreCount <ocaml mutable>: int option;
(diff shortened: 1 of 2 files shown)


semgrep_metrics.atd
```diff
@@ -243,13 +243,20 @@ type supply_chain_config  <ocaml attr="deriving show"> = { ?_rfu: int option; }
...
+    (* tracks unique sessions of the integration *)
     ?sessionId <ocaml mutable>: string option;
-    (* ?? *)
+    (* version of the integration. NOT semgrep version *)
     ?version <ocaml mutable>: string option;
-    (* ?? *)
+    (* type of extension, i.e. vscode, or intellij *)
     ?ty <ocaml mutable>: string option;
+    (* number of autofixes via code actions *)
+    ?autofixCount <ocaml mutable>: int option;
```
brandonspark, line 257: is this autofixes proposed, or autofixes accepted?

semgrep_metrics.atd
```diff
@@ -243,13 +243,20 @@ type supply_chain_config  <ocaml attr="deriving show"> = { ?_rfu: int option; }
...
+    (* how many findings have been ignored *)
+    ?ignoreCount <ocaml mutable>: int option;
```
brandonspark, line 259: also based on the rest of this file we definitely were supposed to use snake case but its fine lol

brandonspark (approved):

aryx (commented):
Thanks for the new comments too!

PR #267, opened 2024-07-01
feat: add rule diagnosis

brandonspark:
This is a stacked diff on top of #272 

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -846,6 +846,75 @@ type cli_output_extra = {
...
+  (* information about the originating pattern in the rule file *)
+  (* This is where the unexpected match came from. *)
+  originating_kind: originating_node_kind;
+  originating_text: snippet;
+  killing_parents: killing_parent list;
+}
+
+type unexpected_no_match_diagnosis = todo
+
+type matching_diagnosis = {
```
aryx, line 912: maybe put that first (in ATD the type definition can be in any order, so better put the high-level type first with a comment explaining its use).

aryx (changes requested):

semgrep_output_v1.atd
aryx: maybe put an example of use, otherwise it's hard to understand all those type definitions and their use.

aryx: Please, move this type at the beginning of this section and add some explanations for what a matching diagnosis is.
I still don't understand what it is and also why it's related to 'semgrep test' 
Also what is its relation to the matching explanation? How it differs?

aryx (changes requested):

semgrep_output_v1.atd
brandonspark: oops, i addressed this and lost the changes accidentally in Git. added back now

semgrep_output_v1.atd
```diff
@@ -868,6 +868,124 @@ type cli_output_extra = {
     ~skipped_rules: skipped_rule list;
 }
 
+(*****************************************************************************)
+(* Matching diagnosis *)
+(*****************************************************************************)
+
```
aryx, line 874: please add a big EXPERIMENTAL comment, so people know they can't rely on it, and it could change.

semgrep_output_v1.atd
```diff
@@ -868,6 +868,124 @@ type cli_output_extra = {
...
+   For instance, suppose we have the rule:
+   1 | all:
+   2 | - pattern: foo(...)
+   3 | - not: foo(good)
+   and the following Python annotated target:
+
+   1 | # ruleid: my_rule
+   2 | foo()
+   3 | # ok: my_rule
+   4 | foo(bad)
```
aryx, line 887: maybe use something like foo(good),  instead of foo(bad).

semgrep_output_v1.atd
```diff
@@ -868,6 +868,124 @@ type cli_output_extra = {
...
+(* Matching diagnosis *)
+(*****************************************************************************)
+
+(* A "matching diagnosis" is a postprocessed interpretation of matching
+   explanations, specific to a particular test-annotated target file.
+
+   For instance, suppose we have the rule:
+   1 | all:
+   2 | - pattern: foo(...)
+   3 | - not: foo(good)
```
aryx, line 881: maybe use foo(goood) here

semgrep_output_v1.atd
aryx: So it's clear that the problem is in the rule, which should be not: foo(good) instead of not: foo(goood).
after all that's the goal of matching diagnosis, to help fix the rule and diagnose.

semgrep_output_v1.atd
```diff
@@ -868,6 +868,124 @@ type cli_output_extra = {
...
+   the form:
+   { matched_text = { line = 4; text = "foo(bad)" };
+     originating_kind = Xpattern;
+     originating_text = { line = 2; text = "- pattern: foo(...)" };
+     killing_parents = [ { killing_parent_kind = Negation;
+                           snippet = { line = 3; text = "- not: foo(good)" } } ]
+   }
+ *)
+type matching_diagnosis = {
+  target: fpath;
```
aryx, line 916: so this is a test target file, maybe add that in comment; it's not any target, it's a test target.

aryx (approved):

PR #268, opened 2024-07-02
feat: as-metavariable

brandonspark:
This PR just adds the `as` field to formulae, due to the addition of embedding matches into arbitrary metavariables.
https://github.com/semgrep/semgrep-proprietary/pull/1730

	  by Semgrep 1.17.0.

### DIFF
--- a/rule_schema_v2.atd
+++ b/rule_schema_v2.atd
@@ -352,6 +352,9 @@ type formula = {
 
   (* NEW: since 1.74 *)
   ?fix: string option;
+
+  (* NEW: since 1.79 *)
+  ?as <ocaml name="as_">: string option;
 }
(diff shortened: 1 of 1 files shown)


IagoAbal (approved):

rule_schema_v2.atd
```diff
@@ -352,6 +352,9 @@ type formula = {
 
   (* NEW: since 1.74 *)
   ?fix: string option;
+
+  (* NEW: since 1.79 *)
+  ?as <ocaml name="as_">: string option;
```
aryx, line 357: maybe a comment saying must be a metavariable would be useful. We actually have an mvar type alias we could
use here instead of string.

PR #272, opened 2024-07-09
feat: matching explanation extra

brandonspark:
## What:
This PR adds an `extra` field to each `matching_explanation` node, which contains kind-specific information, which is not otherwise represented in the tree.

## Why:
I am doing work on "rule diagnosis", which tries to trace back the matching process as well as it can, to determine where a match could have been or was killed.

Unfortunately, this is difficult, because the matching explanation is lossy. We do not store all the intermediary steps of matches in the tree. In particular, for a rule like this:
```
all:
  - pattern: foo($X)
  - inside: |
      def $FUNC(...):
        ...
  - not: foo(bad)
where:
  - metavariable: $X
    regex: "bad2"
```
on a target:
```
def foo():
  foo(abcd)
  foo(bad)
  foo(bad2)
  
foo(abcd)
```
a matching explanation could have this structure:
```
And (2 matches):
  - Xpat "foo($X)" (4 matches)
  - Inside (1 match)
  - Not (1 match)
  - Filter "metavariable-regex" (1 match)
```
We see here that there are a few things wrong:
1) The root `And` node does not have the proper amount of matches assigned to it, having `2` even though the rule actually outputs `1` match.
2) It is difficult to tell how many (or which) matches are the result of the intersection on the positive ranges. This number should be `3`, yet that number appears nowhere in any node in the matching explanations.

This PR aims to fix both issues.

## How:
We have made it such that each root node contains its post-filtering ranges, ensuring that the source of truth for the output ranges at a node is just the matches at that node itself. Without this, you would have to traverse the explanation tree to find the farthest-right `Filter` node, and take its matches.

Doing this displaced the `2 matches` we saw above, which corresponds to the pre-filtering matches of the node in question. We stored that in a new `extra` field, which is on every explanation node. This field is only set in nodes which have a `Filter` child (which is any positive node kind, in practice).

To fix the second issue, we added a `before_negation_matches` field, which is only present for `And` nodes.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -746,7 +746,28 @@ type matching_explanation = {
...
      *)
     loc: location;
-    }
+    (* NEW: since v1.79 *)
+    ?extra: matching_explanation_extra option;
+}
+
+(* For any "extra" information that we cannot fit at the node itself.
+   This is useful for kind-specific information, which we cannot put
+   in the operation itself without giving up our ability to derive `show`.
```
aryx, line 755: I don't understand this deriving show comment.

aryx (approved):
Maybe some of the comments in your PR description could be in the .atd instead.

PR #274, opened 2024-07-16
add semgrep-internal-metavariable-name to rule schema

ihji:
closes CODE-7329

	  by Semgrep 1.17.0.

### DIFF
--- a/rule_schema_v2.atd
+++ b/rule_schema_v2.atd
@@ -410,6 +410,8 @@ type metavariable_cond = {
   ?type_ <json name="type">: string option;
   ?types: string list option;
 
+  ?kind: string option;
+
   (* this covers regex:/pattern:, but also all:/any: with optional where:
    * CHECK: language is valid only when combined with a formula
    * CHECK: constant_propagation: is valid only when combined with regex:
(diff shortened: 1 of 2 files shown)


aryx (approved):
please update also rule_schema_v2.atd and make sure 
semgrep --validate --develop foo.yaml
using the new syntax and this new metavariable-name is working correctly.

rule_schema_v2.atd
```diff
@@ -410,6 +410,8 @@ type metavariable_cond = {
   ?type_ <json name="type">: string option;
   ?types: string list option;
 
+  ?kind: string option;
```
aryx, line 413: please make it an enum, or at least add in comment the kind of strings that are possible here.
Also how do you translate those metavariable-name: in the new syntax? just metavariable_cond with kind?

aryx: Also maybe add a comment before saying it's related to an experimental feature,
metavariable-name: ...

ihji: created a separate PR for updating the comment.

> Also how do you translate those metavariable-name: in the new syntax? just metavariable_cond with kind?

Yes. like metavariable_cond with type for metavariable-type.

PR #279, opened 2024-08-21
+ operators support

Frodan:
Features:
- Support for the assignment operators from Circom language.

	  by Semgrep 1.17.0.


aryx (changes requested):
There is no need to modify those files.
The source of truth is really AST_generic.ml, and you can
leave some Todos in the function that converts from AST_generic to ast_generic_v1.

However, if you make a PR on AST_generic.ml, there is a high chance it will also be denied; we try to keep AST_generic.ml as small as possible and add operators in it when they are present in more than one language. You can use the OtherExpr when converting from circom CST to the generic ASS.

aryx (approved):

PR #280, opened 2024-08-23
feat: add module as possible key

kopecs:
	  by Semgrep 1.17.0.

### DIFF
--- a/rule_schema_v1.yaml
+++ b/rule_schema_v1.yaml
@@ -629,6 +629,9 @@ $defs:
           kind:
             title: Kind keyword
             type: string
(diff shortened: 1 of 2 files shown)


yosefAlsuhaibani (approved):
`module` is indeed now a possible field: lgtm!

aryx (approved):
Can you also update rule_schema_v2.atd ?

kopecs:
Do we use that for anything @aryx? If it is relevant, can you add it to the README/PR checklist?

aryx:
yes it's used in semgrep --validate --develop

PR #282, opened 2024-08-23
Revert "feat: add module as possible key"

kopecs:
Reverts semgrep/semgrep-interfaces#280

Forgot to `make`, and there's a conflict.


kopecs:
Just going to fix in a new PR instead of reverting since this isn't autoapproved.

PR #286, opened 2024-08-31
Extend the definition of 'analyzer' in Semgrep rules

mjambon:
This PR extends the Semgrep rule schema in accordance with the work I did on the OCaml side. The PR is https://github.com/semgrep/semgrep-proprietary/pull/1641 and I'm going to push my work to it very soon.

This PR also adds JSON Schema validation tests. They're useful to check that the JSON Schema is valid, that it validates valid Semgrep rules, and rejects invalid Semgrep rules - without having to rely on another git repo to test this.

test plan: `make test`

This will run `check-jsonschema`. `make setup` will ask the user to install it if it's missing.

	  by Semgrep 1.17.0.


aryx:
Can you also update rule_schema_v2.atd ?

rule_schema_v1.yaml
```diff
@@ -482,13 +514,11 @@ $defs:
...
           analyzer:
-            type: string
-            title: Analyzer to use
+            $ref: "#/$defs/analyzer"
           metavariable:
             type: string
             title: Metavariable to analyze
-          options:
-            type: object
+          additionalProperties: false
```
aryx, line 521: do we want that? Will this create problems down the line when we update the schema and add new fields?
or we can just rely on min-max version to skip the rule?

aryx (approved):

rule_schema_v1.yaml
mjambon: First, I made a mistake here. `additionalProperties:` should not be at the same level as `metavariable`.

Regardless, it's a good question. `additionalProperties` protects the user against spelling mistakes but comes with compatibility issues. I think I'll remove it.

Note that we could check additional fields for spelling mistakes and report an error if they're too close to a missing optional field name. We'd need to be careful to report e.g. `massage` but not `message2` if an optional `message` field isn't specified.

mjambon:
> Can you also update rule_schema_v2.atd ?

I updated it. It's fairly complicated due an ATD adapter that can't be fully generic. It's explained in the comments along with a possible improvement in ATD that would simplify this.

I still have to check if this OCaml code compiles and works (I don't have the impression that we're testing it here).

rule_schema_v2.atd
```diff
@@ -430,11 +430,19 @@ type metavariable_cond = {
 
 type mvar = string
```
aryx, line 432: this TODO comment is still valid? I can see an adapter below.

aryx (approved):

rule_schema_v2.atd
mjambon: good catch, thx

PR #288, opened 2024-09-04
add rpc interface for dependency resolution

bkettle:
Adds the RPC types that will be necessary to perform dependency resolution in OCaml. These changes comprise new types to represent a resolution request, response, and resolved relationships, and an additional optional `id` field on `found_dependency` that is used to associate a dependency with its dependency relationships.

Closes SC-1732.


semgrep_output_v1.atd
```diff
@@ -1110,7 +1110,9 @@ type found_dependency = {
   (* Git ref of the dependency if the dependency comes directly from a git repo.
    * Examples: refs/heads/main, refs/tags/v1.0.0, e5c704df4d308690fed696faf4c86453b4d88a95
    * since 1.66.0 *)
-  ?git_ref: string option
+  ?git_ref: string option;
+  (* ID is used to associate dependencies with each other via depends_on relationships *)
+  ?id: string option;
```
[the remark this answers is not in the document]
bkettle: I've updated the types to no longer use these IDs. Instead, the new version uses the `children` field (of type `dependency_child list`) to encode relationships between dependencies. Ultimately the ID system felt like it required too much out-of-band knowledge about the other `FoundDependency` items, but I was going back and forth between these options for a while so would be curious to hear your thoughts.

semgrep_output_v1.atd
```diff
@@ -1797,12 +1799,44 @@ type output_format = [
   | Emacs
 ]
 
+type manifest = {
```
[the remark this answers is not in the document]
bkettle: I updated this type so that it is now essentially identical to both `manifest_target` in that file and to `manifest` in `Target.ml`. The type is still duplicated here because I wasn't aware of a better way to structure it, though. But hopefully this interface type will only be around for a few months until we get SSC into ocaml entirely

Anyway, curious to hear what you think of this

semgrep_output_v1.atd
```diff
@@ -1817,6 +1862,7 @@ type function_call
      the RPC pipe.
   *)
   | CallValidate of fpath
+  | CallResolveDependencies of resolve_dependencies_params
```
aryx, line 1865: why introduce an intermediate type with a record with just one field?
Why not simpler | CallResolveDependencies of manifest list ?

semgrep_output_v1.atd
```diff
@@ -1827,4 +1873,5 @@ type function_return
   | RetContributions of contributions
   | RetFormatter of string
   | RetValidate of bool
+  | RetResolveDependencies of resolve_dependencies_return
```
aryx, line 1876: Same, I don't think you need the extra intermediate record with just one field ...

aryx (changes requested):

semgrep_output_v1.atd
bkettle: updated!

aryx (approved):

PR #289, opened 2024-09-05
feat: add lockfile path to `found_dependency`

salolivares:
The supply chain team is making several changes to lockfiles/manifest logic in the CLI. In order to simplify some of the works we now require a new field, `lockfile_path` on the `FoundDependency` object.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1104,6 +1104,8 @@ type found_dependency = {
     <ts repr="map">;
   ?resolved_url: string option;
   transitivity: transitivity;
+  (* lockfile_path and line_number point to some location in some file *)
```
aryx, line 1107: add since semgrep 1.87.0 in comment

aryx, line 1107: this comment is not very useful. I'm sure they point to "some" file, would be good to specifiy more precisely what file it is.
Maybe an example at the beginniing of this type would help to understand the type.

aryx (approved):

semgrep_output_v1.atd
salolivares: Added in [1a14725](https://github.com/semgrep/semgrep-interfaces/pull/289/commits/1a14725f4eef424be7b011510b68b17ed6f80787)!

bkettle (approved):

PR #290, opened 2024-09-09
feat(validation): Add simple validate `rpc` as a sibling to `semgrep scan --validate --config`

zzeleznick:
## Description

We want to enable running rules validation in the `OCaml` layer from python instead of using the much slower `jsonschema.validate`. This will also support a more faithful result since it relies on the parsing implementation instead of the rules schema which is more of a reference implementation.

### Planned implementation

```ocaml
let validate (path : Out.fpath) : bool =
  try
    let res = Parse_rule.parse path in
    let valid =
      match res with
      | Ok _ -> true
      | Error _ -> false
    in
    valid
  with
  | _ -> false
```

### Addendum

I also copied the `--config auto` rules to a tmpfile to validate that this RPC was working as expected.

```
time python -c 'from semgrep.rpc_call import validate; import semgrep.semgrep_interfaces.semgrep_output_v1 as out; import time; s=time.time(); print(validate(out.Fpath("/Users/zz/Dev/Semgrep/semgrep-proprietary/cli/tests/default/e2e-other/rules/syntax/bad12.yaml"))); print(time.time() -s)'
```

### Test Plan
- [x] I tested the RPC in a wip osemgrep-pro branch with this change
	  by Semgrep 1.17.0.


nmote (approved):
Seems like a good idea! There is some overhead with the RPC since we spin up a new process each time. There's also serialization overhead but that's not a problem here.

Are you planning to call this repeatedly for each rule, or is the path the path to the config file which has all the rules?

aryx:
I think with all the rules.

semgrep_output_v1.atd
```diff
@@ -1803,6 +1813,7 @@ type function_call
   | CallApplyFixes of apply_fixes_params
   | CallSarifFormat of sarif_format_params
   | CallFormatter of (output_format * cli_output)
+  | CallValidate of validation_params
```
aryx, line 1816: I don't think you need to define an intermediate type when the type is simple no @nmote ?
Here it could be just CallValidate of fpath (* temporary filename containing all the rules *)

semgrep_output_v1.atd
```diff
@@ -1812,4 +1823,5 @@ type function_return
   | RetSarifFormat of sarif_format_return
   | RetContributions of contributions
   | RetFormatter of string
+  | RetValidate of validation_return
```
aryx, line 1826: Same here, just RetValidate of bool (* whether all the rules have valid syntax *)

aryx (changes requested):

semgrep_output_v1.atd
zzeleznick: changed!

semgrep_output_v1.atd
```diff
@@ -1803,6 +1803,7 @@ type function_call
   | CallApplyFixes of apply_fixes_params
   | CallSarifFormat of sarif_format_params
   | CallFormatter of (output_format * cli_output)
+  | CallValidate of fpath
```
aryx, line 1806: please add the comment to say that it's usually a temporary file containing all the rules, probably with JSON
content (we could insteadt have send the rules via a big string through the RPC pipe)

aryx (approved):

PR #292, opened 2024-09-19
Add .plist to list of extensions for Xml

aryx:
Apparently this is a format used by Apple for "property list".
It usually contains XML (but it can also contain binary data
that must be decompressed to xml with the plutil command
(according to chatGPT)

test plan:
make

	  by Semgrep 1.17.0.


kopecs (approved):
Seems OK.

PR #293, opened 2024-09-24
add rpc interface for dump rule partitions

amchiclet:
Closes [SMS-498](https://linear.app/semgrep/issue/SMS-498). Context is in the ticket.

	  by Semgrep 1.17.0.


aryx (approved):
I've expressed my doubt on the distributed scan project, but I'm fine merging this to help experimenting.

PR #294, opened 2024-09-28
Add a metric to record how many interface fields are resolved

ihji:
This PR is for https://github.com/semgrep/semgrep-proprietary/pull/2345

	  by Semgrep 1.17.0.


semgrep_metrics.atd
```diff
@@ -204,13 +204,22 @@ type pro_features = {
...
      *)
     ?numInterfileDiffScanned <ocaml mutable>: (string (* lang *) * int) list
                                               <json repr="object"> option;
+
+    (* The number of interface fields resolved during the Naming_SAST phases.
+     * This is the total count from the naming resolution of all field accesses
+     * in the input source code.
+     * Since semgrep 1.91
+     *)
+    ?numInterfaceVarResolved <ocaml mutable>: int option;
```
aryx, line 221: I don't think we want to collect such metrics. If you're interested in improving our name resolution, just do some benchmarks and extend the metrics in Stat_code.ml (which is used in Graph_code_SIG_SAST.ml for example).

Let's say we collect those metrics, what will you do based on the value?

aryx (changes requested):
I need to be convinved there is a used for those metrics; otherwise not
worth the added complexity.

PR #295, opened 2024-10-07
feat: also allow modules for metavariable-name

kopecs:
Similar to `type` and `types`, extends the existing support for `module` to also allow for `modules`.

	  by Semgrep 1.17.0.


yosefAlsuhaibani (approved):
🙈

rule_schema_v2.atd
```diff
@@ -415,6 +415,7 @@ type metavariable_cond = {
   ?kind: string option;
   (* for semgrep-internal-metavariable-name; consider renaming? for v2 *)
   ?module_ <json name="module">: string option;
+  ?modules: string list option;
```
aryx, line 418: thanks for this one!

PR #296, opened 2024-10-10
Adding Move on Sui to languages

maxmysten:
- [X ] I ran `make setup && make` to update the generated code after editing a `.atd` file (TODO: have a CI check)
- [X ] I made sure we're still backward compatible with old versions of the CLI.
	  by Semgrep 1.17.0.
      I assume that adding a new language is backwards compatiable?

### DIFF
--- a/Language.ml
+++ b/Language.ml
@@ -24,6 +24,7 @@ type t =
 | Kotlin
 | Lisp
 | Lua
+| Move_on_sui
 | Move_on_aptos
 | Ocaml
 | Php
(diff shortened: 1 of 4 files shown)


Language.ml
```diff
@@ -369,6 +370,22 @@ let list = [
...
 };
+(*
+  Move language with SUI flavor
+*)
+{
+  id = Move_on_sui;
+  id_string = "move_on_sui";
+  name = "Move on Sui";
+  keys = [{|move_on_sui|}];
+  exts = [{|.move|}];
```
aryx, line 381: @mjambon will this be problematic that we now have 2 languages (move_on_sui, move_on_aptos) using the same .move extension? Not sure how file targeting will work.

aryx (approved):

PR #298, opened 2024-10-18
chore: Update cli parameters with always fail open

annikalee:
Adds a flag to the cli config to always fail open and suppress errors when set

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1472,6 +1472,8 @@ type engine_configuration = {
     ~generic_slow_rollout: bool;
     (* from 1.63.0 *)
     ?historical_config: historical_configuration option;
+    (* for customers to use the app to configure fail open options *)
```
salolivares, line 1475:
```suggestion
    (* from 1.93 *)
    (* for customers to use the app to configure fail open options *)
```

Let's add the the version!

semgrep_output_v1.atd
```diff
@@ -1472,6 +1472,8 @@ type engine_configuration = {
...
+    ~always_fail_open: bool;
```
salolivares, line 1476: I'm actually unclear on this, perhaps @emjin can chime in. Is there a difference between `~` and `?` ?. I can't really find anything in the docs other than https://atd.readthedocs.io/en/latest/atdgen-tutorial.html#smooth-protocol-upgrades. 

Seems like this should be a safe upgrade and shouldn't break an existing scans when merged in (looks like we default to false if it's not defined).

emjin: `~` is a field with a default value, `?` is an optional field that can be None or Some x

semgrep_output_v1.atd
```diff
@@ -1472,6 +1472,9 @@ type engine_configuration = {
     ~generic_slow_rollout: bool;
     (* from 1.63.0 *)
     ?historical_config: historical_configuration option;
+    (* from 1.93 *)
+    (* for customers to use the app to configure fail open options *)
```
emjin, line 1476:
```suggestion
    (* sent by the app to indicate that fail open should always be enabled, overriding the CLI flag. coupling: server/semgrep_app/saas/models/deployment_products_mixin.py *)
```

emjin: Clarified what this flag does and added a pointer to the comment in semgrep-app

emjin (approved):

annikalee:
@emjin I updated it to always_suppress_errors so asked for a re-review to make sure you're good with the new changes!

PR #299, opened 2024-10-22
feat(ptt): Add space for feature flags to `scan-config` and `scan-info`

SayyantR:
For PTT, in order to ensure that the default lockfile-less logic (to only generate graphs in the absence of a lockfile) is overridden when a customer is using PTT, we need a way to communicate the status of the PTT feature to the CLI from the app. As such, this PR introduces the ability to pass feature flags (`enabled` status only) from the app to the CLI.

	  by Semgrep 1.17.0.


gautambhat (approved):
don't forget to complete the PR checklist :shipit:

PR #300, opened 2024-10-24
Make Core_error.location an option

aryx:
There are cases where we don't have a location for an error,
such as a Fatal error somewhere in semgrep-core. Right now
we use Fpath_.is_fake_file for those cases but it's confusing
then in the error reports so better to not have any location
in those cases

test plan:
see related PR in semgrep

	  by Semgrep 1.17.0.


aryx:
the backward compatiblity warnings are fine; we're modifying the core_output, not the CLI_output.
We can be backward incompattible for semgrep-core, it's fine.

ajbt200128 (approved):

PR #301, opened 2024-10-24
chore(lockfileless): Added rest of `ManifestKinds`

SayyantR:
This PR introduces the rest of the manifest kinds we support to the `ManifestKind` type's possible values.

	  by Semgrep 1.17.0.


bkettle (approved):

semgrep_output_v1.py
```diff
@@ -6548,6 +6548,91 @@ def to_json_string(self, **kw: Any) -> str:
         return json.dumps(self.to_json(), **kw)
 
 
+@dataclass
+class Requirements_:
```
salolivares, line 6552: Huh, wonder why the autogen used an underscore. Perhaps some reserved keyword?

SayyantR: Looks like `Requirements` already exists in that file: https://github.com/semgrep/semgrep-interfaces/blob/5149397e4cda142518a2d582a1ca9fe4bbbd5e54/semgrep_output_v1.py#L3935

SayyantR: Should we rename to something different? Maybe `RequirementsManifestKind`...? Though I don't like the sound of that haha

SayyantR: Oop i'll keep it as is then haha

bkettle: Should we do `RequirementsIn` actually? Since that's the format we require? Keep with the other ones where we specify their extension

salolivares: Our file matcher for requirements is called `PipRequirementsMatcher`. Could do that, but I think it's fine as is. It might be confusing for someone in python land, but not a blocker.

SayyantR: I think I'll opt for `PipRequirements` since it's specifically for that build system- preciate the suggestions!

SayyantR: Context: https://semgrepinc.slack.com/archives/C05GV9WR2MB/p1729808284289329

We decided on `RequirementsIn` with the option to pivot if we expand our range of manifest files for `pip` later

PR #302, opened 2024-10-24
chore(dep-resolution): add dependency resolution errors to core error type

bkettle:
As part of the ongoing lockfileless work, we aim to integrate errors that occur during dependency resolution into the error framework that the rest of the CLI uses. This PR adds dependency resolution errors as an error type.

Dependency _parser_ errors remain as their own type for now, but in the future may be integrated as well by adding them as a variant of `DependencyResolutionError`.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -511,6 +511,7 @@ type error_type
    * *)
   | PatternParseError0 <json name="Pattern parse error">
   | IncompatibleRule0 <json name="Incompatible rule">
+  | DependencyResolutionError of resolution_error
```
salolivares, line 514: wdyt about adding to the documentation to make it clear what version we introduced this error!

salolivares (approved):

semgrep_output_v1.atd
aryx: yes, we usually do that, like (* since 1.94.0 *).

aryx (approved):

semgrep_output_v1.atd
bkettle: updated!

PR #305, opened 2024-10-30
chore(sca): add comments to manifest kinds

bkettle:
In order to consolidate the `manifest_kind` type in interfaces with `Manifest_kind.ml`, this PR:
- transfers the comments from the ML file to interfaces
- adds `eq`, `show`, and `yojson` derivations to the ATD definition

	  by Semgrep 1.17.0.

### DIFF
--- a/semgrep_output_v1.atd
+++ b/semgrep_output_v1.atd
@@ -1824,22 +1824,34 @@ type output_format = [
   | Emacs
 ]
 
-(* TODO: merge with Manifest_kind.ml *)
-type manifest_kind <python decorator="dataclass(frozen=True)"> = [
-  | RequirementsIn (* Manifest file for pip *)
+type manifest_kind 
+  <ocaml attr="deriving show">
+  <python decorator="dataclass(frozen=True)"> 
+  = [
+| RequirementsIn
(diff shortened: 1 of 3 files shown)


salolivares (approved):
lgtm!

semgrep_output_v1.atd
```diff
@@ -1824,22 +1824,33 @@ type output_format = [
   | Emacs
 ]
 
-(* TODO: merge with Manifest_kind.ml *)
-type manifest_kind <python decorator="dataclass(frozen=True)"> = [
-  | RequirementsIn (* Manifest file for pip *)
+type manifest_kind
```
aryx, line 1827: @mjambon what do we need to add to generate proper Variant like RequirementsIn instead of the `RequirementsIn in the generated OCaml code?

aryx (commented):
nice

aryx:
Ideally we would also use those types in the pysemgrep code.

semgrep_output_v1.atd
bkettle: Hmm, it looks like that might even be supported today with `<ocaml repr="classic">`: https://atd.readthedocs.io/en/latest/atdgen.html#sum-types

Do you think I should switch this one over?

aryx: I prefer yes. They are simpler to handle.

PR #306, opened 2024-10-31
feat(sms): add scm_run_id to scan-create interface

chmccreery:
For SMS, we would like to be able to connect a particular Scan object to the SCM (source control manager) run_id that it is associated with. Details [here](https://www.notion.so/semgrep/SMS-Supportability-1273009241a880579e9ae90928391a0e?pvs=4#1313009241a880d79d77eb0d6773ad56).

If we create the SCM run before the scan exists (desirable because there are actually many steps between us receiving a webhook and a Scan object getting created, leading to a significant time delay), then when we create the Scan, we need to know what the SCM run id is in order to connect it to the Scan and update it as the Scan progresses. This interface change allows us to pass the SCM run from the CLI to the backend. 

Changes in `semgrep` to follow this PR will actually set this value based on an environment variable and then send it to the app.

	  by Semgrep 1.17.0.


amchiclet:
We're discussing more context on this offline. Will come back to review once things are clearer, thanks!

aryx (approved):
Can you add a comment explaining what SCM mean, what are example of values to make things more concrete (it is like "git" or "mercurial" thing?) and vaguely why we need it to better understand the field.
Also 1.95.0 was out yesterday so you probably want 1.96.0

amchiclet (approved):
Thanks!

Optional: in my mind, I think adding just a bit more context in the PR description through a concrete github example helped me understand this issue.

`For SMS, we would like to be able to connect a particular Scan object to the SCM (source control manager) run_id that it is associated with, e.g. so that we can tell GitHub (the SCM) that a PR check (which runs a scan and has a Scan object) is running or completed.`

PR #309, opened 2024-11-05
Extend AWS support for temporary tokens [SRTSR-47]

LewisArdern:
This updates interfaces for supporting temporary tokens in AWS.

Tested this locally and it seems to enforce it if providing a list vs a string:

```
semgrep error: Invalid rule schema
['list1', 'list2'] is not of type 'string'
```

General checklist doesn't apply since we aren't updating an .atd for this.


kopecs (approved):
lgtm

PR #310, opened 2024-11-06
Add format_context and complete output_format for RPC call to format

aryx:
This will help SAF-1696

test plan:
see related PR in semgrep

	  by Semgrep 1.17.0.


aryx:
the error concerns function_call which is part of the RPC framework which does not have to be backward compatible so we're fine.

IagoAbal:
> the error concerns function_call which is part of the RPC framework which does not have to be backward compatible so we're fine.

It would be nice to add `function_call` to a whitelist or something.

aryx:
@mjambon can we customize the backward compatiblity check and do the whitelisting iago proposed?

PR #311, opened 2024-11-07
chore(lfs): Added `manifest_path` to `FoundDependency`

SayyantR:
In order to better support PTT, the CLI will now be able to record a dependency's manifest filepath using the new attribute added in this PR.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1113,6 +1113,11 @@ type found_dependency = {
     <ts repr="map">;
   ?resolved_url: string option;
   transitivity: transitivity;
+  (* Path to the manifest file which the lockfile can be associated with
```
bkettle, line 1116: Maybe "Path to the manifest file that defines the project containing this dependency"?

bkettle (approved):

PR #312, opened 2024-11-07
chore(dep-resolution): update dep-resolution RPC interface to support lockfiles

bkettle:
Updates the RPC interface used to resolve dependencies in OCaml to support
lockfiles as well as manifests. This will allow us to begin porting lockfile
parsers to Ocaml as we move forward.

These changes are _not_ backwards compatible. However, they only affect types
that are used within the CLI, from the Python code to the OCaml code, so as long
as we are careful with releasing the changes to the CLI, it should be safe.

	  by Semgrep 1.17.0.


bkettle:
@aryx I have these changes I want to make to the RPC interface (not quite ready to merge yet). They aren't backwards compatible, so the check fails. As far as I can tell, though, there is no real need for backwards compatibility in these RPC types (if ATD had a module system so that we could use types from `semgrep-interfaces` I think the RPC types would be their own ATD file like `Input_to_core` etc.).

Can we ignore the backwards compatibility check here? Or should I restructure these changes to be backwards compatible? (I think the latter would probably look something like adding a new RPC call `ResolveDependenciesV2`)

semgrep_output_v1.atd
```diff
@@ -71,7 +71,7 @@ type raw_json <ocaml module="Yojson.Basic" t="t"> = abstract
 (*****************************************************************************)
 
 (* File path. less: could convert directly to Path class of pathlib library for Python *)
-type fpath <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True)">= string wrap <ocaml module="ATD_string_wrap.Fpath">
+type fpath <ocaml attr="deriving show, eq, yojson"> <python decorator="dataclass(frozen=True)">= string wrap <ocaml module="ATD_string_wrap.Fpath">
```
aryx, line 74: Why do you need yojson? In theory if you need to generate json strings from those types you can use
the generated string_of_xxx

aryx (approved):

semgrep_output_v1.atd
```diff
@@ -1852,6 +1852,27 @@ type format_context
...
+    | NpmPackageLockJson
+    | YarnLock
+    | PnpmLock
+    | GemfileLock
+    | GoMod
+    | CargoLock
+    | MavenDepTree (* Not a real lockfile *)
+    | GradleLockfile
+    | ComposerLock
+    | NugetPackageLockJson
```
gautambhat, line 1870: Fwiw, this would be more accurate if it's `NugetPackagesLockJson` because the lockfile is `packages.lock.json`

bkettle: oh, great, thanks. will fix

bkettle: fixed!

semgrep_output_v1.atd
bkettle: Turns out that it was unnecessary, removed!

bkettle:
@aryx I'm finally getting ready to merge this and the accompanying `semgrep-proprietary` PR, but it looks like I can't do so because the backwards compatibility check is required. Is this something you can help with?

aryx:
Yes I can override the backward compatibility check because I'm admin on this repo I think. @mjambon too.

aryx:
@mjambon we need to be able to whitelist a few entry types to not be concerned by the backward compatibility check (e.g., the types for semgrep-core or here the types for Nat's RPC, or we need a module system so we can use
different files for those different purpose.

aryx:
The whitelist seems the easiest option.

bkettle:
Thanks @aryx ! Agree that a whitelist seems nice; it would be good to be able to make changes to the RPC interfaces more easily.

aryx:
yes, and merging the semgrep-interface inside the semgrep-pro repo would accelerate things even more.
That's one of the task of @mjambon for Q4 actually.

PR #313, opened 2024-11-12
feat(sca): new SCA stats schema

salolivares:
# Context

The [existing "bytes parsed" metric](https://metabase.corp.semgrep.dev/dashboard/272-ssc-adoption-dashboard?tab=147-summary) we are currently relying on provides a limited perspective on customer SSC adoption, as it aggregates usage across languages without fully capturing which languages are meaningfully used. 

We want to augment the stats schema to improve metrics around Java dependency detection, lockfile-less scanning, and python requirements identification.

	  by Semgrep 1.17.0.


semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,42 @@ type ci_scan_complete_stats = {
...
+  (* since 1.98.0 *)
+  (* In collaboration with the Data Science team, it was suggested
+   * that we start to group stats by product for organizational purposes.
+   *)
+  ?supply_chain_stats: supply_chain_stats option;
+}
+
+(* Valid schema versions. Update when making breaking changes. *)
+type stats_schema_version <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True)"> = [
+  | V1_0_0 <json name="1.0.0"> (* initial version *)
```
salolivares, line 1699: We need a schema version for data warehousing purposes down stream. 

A couple worries:
- We forget to update the schema version on changes
- We'll only ever add a new schema version on breaking changes, so maybe semantic versioning doesn't make sense? Perhaps just a date (`20241112`) would suffice?

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,42 @@ type ci_scan_complete_stats = {
...
+
+type subproject_stats = {
+  dependency_source_files: string list;
+  (* Results of dependency resolution, null if resolution failed *)
+  resolved_stats: dependency_resolution_stats nullable;
+}
+
+type supply_chain_stats = {
+  subproject_stats: subproject_stats list;
+  lockfile_scan_info: (string * int) list
```
salolivares, line 1721: I'm planning to move `lockfile_scan_info` into the supply chain stats object. However considering it's deprecated, does that even make sense? Should I just leave it out?

semgrep_output_v1.atd
bkettle: Agree that there seems to not be much reason to use semantic versioning here. I like the idea of using a date.

In general, though, we don't allow backwards-incompatible changes to the interface types. What defines a breaking change here? Would something like adding a field require updating the version?

semgrep_output_v1.atd
bkettle: I think it seems duplicative of the info in `subproejct_stats` -- I think I would vote that we just leave it where it is for now, maybe

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,42 @@ type ci_scan_complete_stats = {
...
+type dependency_resolution_stats = {
+  resolution_method: resolution_method;
+  dependency_count: int;
+  ecosystem: ecosystem;
+}
+
+type subproject_stats = {
+  dependency_source_files: string list;
+  (* Results of dependency resolution, null if resolution failed *)
+  resolved_stats: dependency_resolution_stats nullable;
```
bkettle, line 1716: why nullable here instead of option?

salolivares: No explicit reason. I was working of this sample json:
```
{
    "subproject_stats": [
        ...
        {
            "dependency_sources": ["proj_c/gradle.kts"],
            "resolved_stats": null
        }
    ]
}
```

In this case, nullable makes it clear that `resolved_stats` explicity has no value. In the end, i don't think it has impact on the downstream metrics -- just on how we use it!

bkettle: Ah, looks like `nullable` in ATD still becomes an `option` in Ocaml - seems solid then! https://atd.readthedocs.io/en/latest/atdgen.html#default-type-mapping

semgrep_output_v1.atd
salolivares: > Agree that there seems to not be much reason to use semantic versioning here. I like the idea of using a date.

I'll switch to a date then!

>  Would something like adding a field require updating the version? 

No, adding a field would not require updating the version. Basically anytime we remove/change an existing field (so basically anytime the backward compatibility PR check fails). I imagine we won't have too many backward incompatible changes due to the enforcement of backward compatibility for the last X versions. 

Now that I think about it, there’s a case for not restricting version updates solely to breaking changes. It’s really up to us, since the schema version primarily serves to inform the data warehouse (Snowflake) on how to process the incoming stats JSON. 

For instance, if we add a new field and deprecate an old one -- like if `subproject_stats` isn’t sufficient a year from now and we introduce `subproject_stats_v2` -- updating the schema version could guide downstream processing to start using `subproject_stats_v2`. We’d then update the schema version again once we remove `subproject_stats`. 

So, updating on additions or deprecations might make sense to keep data handling in sync.

semgrep_output_v1.atd
salolivares: Perfect! I'll leave as is then!

semgrep_output_v1.atd
salolivares: @bkettle changed to:
```
type stats_schema_version <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True)"> = [
  | SSV20241112 (* initial version *)
] 
```

Prefixed with `SSV` since it can't start with a number.

semgrep_output_v1.atd
salolivares: Removed in [0fbb591](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/0fbb591f12b76200f26aa80bc09ea95c43468fc9)

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+   *)
+  ?supply_chain_stats: supply_chain_stats option;
+}
+
+(* Valid schema versions. Update when making breaking changes. *)
+type stats_schema_version <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True)"> = [
+  | SSV20241112 (* initial version *)
+] 
+
+type resolution_method <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True, order=True)"> = [
```
salolivares, line 1702: @bkettle we have two columns, `lockfile_flag` and `manifest_flag`, in the [proposed SSC fact table](https://docs.google.com/spreadsheets/d/14jNMAvE2YRrnpmRPzodIIjohM9s9iLUO/edit?gid=1327685153#gid=1327685153). 

I considered adding `ManifestParsing` to specify whether we parsed a lockfile or a manifest file, but that may be redundant.

For tracking, Khoa wants to distinguish `dependency_source_files` as either lockfiles or manifest files. Currently, we’re using `LockfileParsing` or `DynamicResolution`. It seems we can map `LockfileParsing` to `lockfile_flag` and `DynamicResolution` to `manifest_flag` (since lockfileless setups depend on some kind of manifest file, correct?).

bkettle: Lockfileless setups might not depend on a manifest file in the future---e.g. for Go, there is only a `go.mod` file that serves both as a manifest and a lockfile. We will probably categorize this as a lockfile, but PTT would require it be resolved with dynamic resolution. Might not be worth trying to build for that now, though.

What's the purpose of the `lockfile_flag` vs `manifest_flag`?

I wonder if we should consider using the new `dependency_source` type to be introduced in https://github.com/semgrep/semgrep-interfaces/pull/312 in the stats JSON?

salolivares: > Lockfileless setups might not depend on a manifest file in the future

Makes sense, thanks for the context

> What's the purpose of the lockfile_flag vs manifest_flag?

I believe it's just used for tracking purposes -- just to have more metadata available. 

> I wonder if we should consider using the new dependency_source type to be introduced in https://github.com/semgrep/semgrep-interfaces/pull/312 in the stats JSON?

Ohhh. that would be helpful actually.

bkettle: OK, there is a bit more work to do there because that PR needs to be merged in sync with the `semgrep-pro` PR, and there are a couple things left to tweak in the `semgrep-pro` PR, but I am hoping to make it happen tomorrow or the day after.

salolivares: perfect, I have plenty to do in the meantime!

semgrep_output_v1.atd
```diff
@@ -1647,7 +1647,12 @@ type ci_scan_complete = {
   ?final_attempt: bool option; (* always optional *)
   }
 
+(* Update the stats_schema_version if you are changing this type.
```
aryx, line 1650: I'm not sure I understand the goal of this schema_version; We can't have breaking change ... we must support older versions of Semgrep, which is why for instance you've added this new field as a `?` because otherwise we would not parse correctly the output of semgrep 1.95.0 for example.
We have actually a window of versions we support (@chmccreery is actually working on moving up the minimum version to 1.50.0 or something) but in any case what matters is this mimimal version of semgrep. So we already send the semgrep version in another field; we don't need another version scheme.

aryx (changes requested):

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+] 
+
+type resolution_method <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True, order=True)"> = [
+  | LockfileParsing 
+  | DynamicResolution 
+]
+
+type dependency_resolution_stats = {
+  resolution_method: resolution_method;
+  dependency_count: int;
```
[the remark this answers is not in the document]
salolivares: Correct. It doesn't matter what resolution method is used, the whole point is to figure out which ecosystem and how many dependencies there are in a subproject.

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+
+type subproject_stats = {
+  (* Files used to determine the subproject's dependencies (lockfiles, manifest files, etc) *)
+  dependency_source_files: string list;
+  (* Results of dependency resolution, null if resolution failed *)
+  resolved_stats: dependency_resolution_stats nullable;
+}
+
+type supply_chain_stats = {
+  subproject_stats: subproject_stats list;
```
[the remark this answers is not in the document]
salolivares: Good call out. done in [8cc857e](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/8cc857ee19cebdc4b7ddf617d942b3ab757864d8)

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+type dependency_resolution_stats = {
+  resolution_method: resolution_method;
+  dependency_count: int;
+  ecosystem: ecosystem;
+}
+
+type subproject_stats = {
+  (* Files used to determine the subproject's dependencies (lockfiles, manifest files, etc) *)
+  dependency_source_files: string list;
+  (* Results of dependency resolution, null if resolution failed *)
```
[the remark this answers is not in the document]
salolivares: Good to know! Updated to option in [8cc857e](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/8cc857ee19cebdc4b7ddf617d942b3ab757864d8)

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+
+type dependency_resolution_stats = {
+  resolution_method: resolution_method;
+  dependency_count: int;
+  ecosystem: ecosystem;
+}
+
+type subproject_stats = {
+  (* Files used to determine the subproject's dependencies (lockfiles, manifest files, etc) *)
+  dependency_source_files: string list;
```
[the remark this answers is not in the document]
salolivares: Oops. Updated in [8cc857e](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/8cc857ee19cebdc4b7ddf617d942b3ab757864d8)

semgrep_output_v1.atd
salolivares: We can’t have breaking changes — that’s totally clear. The schema version idea is more about making life easier for downstream tasks, especially in Snowflake.

Right now, the stats JSON gets transformed into a stats table. If we ever evolve the schema — like adding fields —it’d be really handy to have a version marker that tells Snowflake tasks, “Hey, this is the structure you’re dealing with.” We can definitely check for the existence of new keys instead, but having a schema version simplifies things a lot, especially when working with historical data. 

For the Data Science team, it means they don’t have to guess when fields were added or rework transformation - it’s just cleaner and easier to manage. It’s kind of a future-proofing move. The schema evolves, nothing breaks, and downstream logic stays sane. 

That said, now that I think about it, perhaps when doing the data science bit, we can join with the scan table and grab the semgrep_version from the meta column. We could use semgrep_version in place of this proposed schema version. Or maybe just bake the semgrep verison into stats? What do you think?

semgrep_output_v1.atd
salolivares: @bkettle using new dependency type here [fb81785](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/fb81785f2e43ad0c3166602090a43e0e8c044278)

semgrep_output_v1.atd
```diff
@@ -1656,6 +1661,7 @@ type ci_scan_complete_stats = {
     <json repr="object">
     <python repr="dict">
     <ts repr="map">;
+  (* Deprecated - subproject_stats is used instead *)
```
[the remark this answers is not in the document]
salolivares: I've removed the comment to reflect reality. Yes, we're still generating the data, i'll mark as deprecated once we stop populating that data.

semgrep_output_v1.atd
salolivares: @aryx left some good feedback related to schema version. might end up keying off of semgrep version instead. see response below https://github.com/semgrep/semgrep-interfaces/pull/313#discussion_r1844477512

bkettle: Yeah, I think that seems like a good idea

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+  ?supply_chain_stats: supply_chain_stats option;
+}
+
+(* Valid schema versions. Update when making breaking changes. *)
+type stats_schema_version <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True)"> = [
+  | SSV20241112 (* initial version *)
+] 
+
+type resolution_method <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True, order=True)"> = [
+  | LockfileParsing
```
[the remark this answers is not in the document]
salolivares: TIL! good to keep in mind.

semgrep_output_v1.atd
```diff
@@ -1680,6 +1686,39 @@ type ci_scan_complete_stats = {
...
+  (* since 1.98.0 *)
+  (* In collaboration with the Data Science team, it was suggested
+   * that we start to group stats by product for organizational purposes.
+   *)
+  ?supply_chain_stats: supply_chain_stats option;
+}
+
+(* Valid schema versions. Update when making breaking changes. *)
+type stats_schema_version <ocaml attr="deriving show"> <python decorator="dataclass(frozen=True)"> = [
+  | SSV20241112 (* initial version *)
```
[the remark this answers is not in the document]
salolivares: See: https://github.com/semgrep/semgrep-interfaces/pull/313#discussion_r1844477512

semgrep_output_v1.atd
```diff
@@ -1680,6 +1680,46 @@ type ci_scan_complete_stats = {
...
+  (* Unique identifier for the subproject *)
+  subproject_id: string;
+  (* Files used to determine the subproject's dependencies (lockfiles, manifest files, etc) *)
+  dependency_sources: dependency_source_file list;
+  (* Results of dependency resolution, null if resolution failed *)
+  ?resolved_stats: dependency_resolution_stats option;
+}
+
+type supply_chain_stats = {
+  subprojects_stats: subproject_stats list;
```
chmccreery, line 1722: For projects without supply chain, would this just be an empty list?

semgrep_output_v1.atd
```diff
@@ -1680,6 +1680,46 @@ type ci_scan_complete_stats = {
     <python repr="dict">
     <ts repr="map">
     option;
+
+  (* since 1.98.0 *)
+  (* In collaboration with the Data Science team, it was suggested
+   * that we start to group stats by product for organizational purposes.
+   *)
+  ?supply_chain_stats: supply_chain_stats option;
```
chmccreery, line 1688: FYI, now that our synthetic tests are using the atd-defined stuff, when we add fields we will have to add them to the synthetic tests as well. SMS team can handle this, but please continue to keep us in the loop - specifically on PR's that bump semgrep-interfaces versions within semgrep-app.

semgrep_output_v1.atd
salolivares: I was thinking since `supply_chain_stats` is optional I would leave the fields as `None` or empty. 

But if preferred, I can define it for all scans and in that case, yes, `subprojects_stats` would be an empty list.

semgrep_output_v1.atd
salolivares: Gotcha! Thanks for the heads up @chmccreery.

semgrep_output_v1.atd
chmccreery: Yeah, I like using the `semgrep version` for this, but don't feel too strongly. It just seems like another thing to try to keep in sync manually, so another opportunity for human-introduced errors.

aryx (approved):
LGTM

semgrep_output_v1.atd
```diff
@@ -1680,6 +1680,46 @@ type ci_scan_complete_stats = {
     <python repr="dict">
     <ts repr="map">
     option;
+
+  (* since 1.98.0 *)
+  (* In collaboration with the Data Science team, it was suggested
```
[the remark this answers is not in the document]
chmccreery: Oh, ok, I misunderstood the data science piece. Doesn't this only relaly help them if these are in a DB somewhere?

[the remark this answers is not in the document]
salolivares: Ah! As part of [my project](https://www.notion.so/semgrep/Improved-SSC-Adoption-Metrics-1303009241a8808d84f8feade1a7cd62?pvs=4), the stats in s3 are being piped into snowflake. 

The s3 to snowflake sync is already happening for staging:
image

The stats json is being transformed into a nice table to derive metrics from!

semgrep_output_v1.atd
```diff
@@ -1680,6 +1680,46 @@ type ci_scan_complete_stats = {
...
+] 
+
+type dependency_source_file = {
+  kind: dependency_source_file_kind;
+  path: fpath;
+}
+
+type subproject_stats = {
+  (* Unique identifier for the subproject *)
+  subproject_id: string;
```
bkettle, line 1714: Might be worth describing the scope within which this ID will be unique - if we do the hash-of-all-dependency-source-files then it should be consistent across scans and unique per-repo, right?

semgrep_output_v1.atd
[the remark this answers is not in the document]
salolivares: I’m envisioning a future where we no longer store scan stats in our PostgreSQL database. Instead, the stats would be piped directly to S3 and then to Snowflake. This approach would eliminate the need for a JSON column in the scan table. however careful planning is still required, especially since some folks may rely on that column for other purposes.

More context: after discussions with the infra team, we’re increasingly moving towards using Snowflake to centralize all our data. Snowflake’s power make it much better suited for data engineering tasks. For this project, we plan to create several derived tables from the new stats table in Snowflake. 

These tables will be made available in metabase so we can create some dashboards! Essentially, Snowflake will act as our powerful backend, while metabase serves as the frontend, displaying dashboards powered by snowflake’s data.

>  If so, can we make placeholders for the other products as well and follow the same path that Jonas laid out for moving the meta field in ScanRequest to more structured types in project_metadata and scan_metadata.

Placeholders for other products isn't a bad idea. In fact @khoa-semgrep suggested that it would be helpful to eventually divide the stats key into product-specific stats and generic stats applicable across all products

semgrep_output_v1.atd
salolivares: > then it should be consistent across scans and unique per-repo, right?

Hmmmm, for the most part it should be consistent across scans. The only problem is when a dependency source file is added, renamed, or removed. But I think that's okay since changes are infrequent?

salolivares: Guess you could also make the argument different set of dependency source files == different subproject.

semgrep_output_v1.atd
```diff
@@ -1680,6 +1680,46 @@ type ci_scan_complete_stats = {
...
+  (* Files used to determine the subproject's dependencies (lockfiles, manifest files, etc) *)
+  dependency_sources: dependency_source_file list;
+  (* Results of dependency resolution, null if resolution failed *)
```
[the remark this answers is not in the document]
salolivares: good catch, updated in [c2cf640](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/c2cf6408bf4af4ac7221a9c6e46d4da41622bdaa)

semgrep_output_v1.atd
salolivares: updated in [c2cf640](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/c2cf6408bf4af4ac7221a9c6e46d4da41622bdaa)

semgrep_output_v1.atd
salolivares: Updated in [c2cf640](https://github.com/semgrep/semgrep-interfaces/pull/313/commits/c2cf6408bf4af4ac7221a9c6e46d4da41622bdaa)!

bkettle (approved):

semgrep_output_v1.atd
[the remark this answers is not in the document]
salolivares: It’s worth noting that this is the tentative plan so far (having S3 stats piped to Snowflake). 

The staging test is working well, but if the larger volume of production stats becomes too costly, the backup plan is to revert to using the stats column. We don’t anticipate SSC stats becoming as large as the errors were, but I’d still like to avoid that scenario.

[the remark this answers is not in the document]
chmccreery: > The stats json is being transformed into a nice table to derive metrics from!

Oh, that's very nice! Ok, this seems fine. 

> I’m envisioning a future where we no longer store scan stats in our PostgreSQL database.

FWIW we are depending on several fields that are in the stats field now in the app, so we can't just move it to S3 🙁 In my ideal world, rather than living in JSON, these fields that have simple, size-reasonable data types (bools, ints, floats) would each have their own column in the scan table and only the longer and unbounded-in-size fields would be in S3 (e.g. "errors"). We are already doing that for "errors", but they're being stored to S3 as part of the scan pipeline S3. It may make sense to combine our scan-pipeline buckets (configs, results, stats with errors) with whatever buckets you're using now. The old ones expire after a few weeks, which isn't great. The ones I'm referring to are [here](https://us-west-2.console.aws.amazon.com/s3/buckets/scan-pipeline-prod?region=us-west-2&bucketType=general&tab=objects) and were originally created to support the RFS work.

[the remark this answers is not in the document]
salolivares: > FWIW we are depending on several fields that are in the stats field now in the app, so we can't just move it to S3 🙁

Ack! I figured some folks might rely on it. 

> these fields that have simple, size-reasonable data types (bools, ints, floats) would each have their own column in the scan table and only the longer and unbounded-in-size fields would be in S3 (e.g. "errors")

that makes sense to me! 

> It may make sense to combine our scan-pipeline buckets (configs, results, stats with errors) with whatever buckets you're using now.

I would love to combine them. At the moment, I went with a [new bucket](https://us-west-2.console.aws.amazon.com/s3/buckets/semgrep-scan-stats-prod?region=us-west-2&bucketType=general&tab=objects) to hold _just_ stats in a very specific format to make ingest easier for snowflake. 

Lots to think about still! @chmccreery appreciate discussion, i'll be sure to capture this in a retro doc for future consideration.

PR #318, opened 2024-12-02
Pass format_context also to RPC CallSarifFormat

aryx:
This will enable to gate some fields with is_logged_in

test plan:
see related PR in semgrep-pro

	  by Semgrep 1.17.0.


aryx:
It's not backward compatible but it's fine; this is RPC stuff so it does not have to be backward compatible
like the regular JSON output of the CLI (or our comm with semgrep ci)

amchiclet (approved):

PR #320, opened 2024-12-02
feat(sca): add conan (C++) ecosystem

salolivares:
This PR adds the Conan C++ package manager to the ecosystem type.

	  by Semgrep 1.17.0.


aaronmichaelacosta (approved):
Assuming you follow the correct steps after making the change, it LGTM

semgrep_output_v1.atd
```diff
@@ -1091,6 +1091,7 @@ type ecosystem <python decorator="dataclass(frozen=True)"> <ocaml attr="deriving
   (* Deprecated: Mix is a build system, should use Hex, which is the ecosystem *)
   | Mix <json name="mix">
   | Hex <json name="hex">
+  | Conan <json name="conan">
```
bkettle, line 1094: I wonder if this is the best approach here. It may be, but it also feels slightly awkward because we don't technically support this ecosystem from a SCA rules standpoint yet (and C++ is weird enough that I personally am not super confident that Conan really fits our idea of an "ecosystem"). 

For package managers that we don't yet support, I wonder if it would make more sense to add an unknown ecoosystem or, probably better, allow ecosystem to be optional in the places where we convert from lockfile kind to ecosystem. What do you think?

salolivares:
closing. need a larger discussion on what a C++ ecosystem should look like.

semgrep_output_v1.atd
bkettle: > That said, we’ll likely need to support C++ package managers in the future, and when that happens, we’ll still need a way to define an overarching ecosystem. Following the Python approach (Pypi), we could introduce something like CPlusPlus to represent the broader ecosystem (assuming ++ isn’t allowed in the type name).

Yeah, I think this will require some thought when we get there. It feels odd because while there are different package managers for python, AFAIK they all use the same basic registry definition defined by Python, but that's not the case for C++ (there is no such concept of a registry in the language spec afaik). This sounds reasonable for now and for the general case of unsupported subproject types.

PR #326, opened 2024-12-05
Include new `app_blocking_match_based_ids` field in `ci_scan_complete_response`

mmcqd:
Right now whether or not a CI scan should exit with code 1 is determined _mostly_ by the CLI, but the app can use the `app_block_override` field of `ci_scan_complete_response` to tell the CLI to exit with code 1 no matter what. This is currently used to support blocking scans when we detect a dependency with a blocked license.

We're currently working on a project which includes making it more customizable within the app whether or not a given finding should be blocking, right now just for supply chain, but with the aim to eventually move the computation of blocking findings entirely to the app. So we'd like to be able to tell the CLI not only "please block" but also "here are the list of findings that the app determined are blocking".

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -1617,12 +1617,15 @@ type parsing_stats = {
...
 
+
 (* Response by the backend to the CLI to the POST /complete *)
 type ci_scan_complete_response <ocaml attr="deriving show"> = {
    success: bool;
    ~app_block_override: bool;
    (* only when app_block_override is true *)
    ~app_block_reason: string;
+   (* match_based_ids of findings that semgrep-app determined should cause the scan to block *)
+   ~app_blocking_match_based_ids : string list;
```
aryx, line 1628: we should define a separate string wrap type for match_based_id, just like we have a special
type for fpath, uri, rule_id, etc. It's not good to use string for everything.

chmccreery:
Sorry, going to leave some high-level feedback here since I don't have context on the project. Feel free to ping me in Slack or put time on my calendar if you want to discuss in person:

How would we know ahead of time which match_based_id's are blocking? We don't know a match_based_id for a particular finding until we've seen that finding at least once - and we don't know the various qualities of that finding until it's been detected at least once, so don't know whether or not to block on it. Seems like a chicken and egg problem, but maybe I'm missing something?

mmcqd:
Discussed this with @chmccreery on a call, we got the go ahead

semgrep_output_v1.atd
mmcqd: sure

mmcqd: hm actually, doing that seems not super useful since there's no special internal representation for them. They're just strings and all we ever do with them is compare for equality

chmccreery (approved):

semgrep_output_v1.atd
aryx: It does not matter. Filenames are also strings, but we still want a different type for them. Otherwise they will be nothing that prevents a filename to go inside a match_based_id.
Types!

mmcqd: ok yeah I agree

PR #329, opened 2024-12-10
chore(ast-generic): Add `Cls` variant

brandonspark:
	  generated by Semgrep 1.50.0.


yosefAlsuhaibani (approved):

aryx:
you should not care about ast_generic_v1.atd.
Nobody is using it.
You can put a raise Todo for Cls during the conversion. It would not matter.

yosefAlsuhaibani:
If it's unused, should we delete it?

aryx:
we should

PR #330, opened 2024-12-16
feat: added cocoapods lockfile to the list of supported lockfiles

aaronmichaelacosta:
Added necessary interface changes to support 1) a new ecosystem, cocoapods, 2) new lockfile + manifest types for cocoapods, and 3) a new parser type for (you guessed it) cocoapods.

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -509,6 +509,7 @@ type ecosystem
   | Nuget <json name="nuget">
   | Pub <json name="pub">
   | SwiftPM <json name="swiftpm">
+  | Cocoapods <json name="cocoapods">
```
bkettle, line 512: Your understanding is that SwiftPM and Cocoapods are entirely separate package namespaces, right? I did a bit of searching just now and that seems to be the case, but it also seems like they might both use git URLs as the primary identifier. I just want to ensure that this isn't an npm/yarn or pip/poetry situation

bkettle (approved):

salolivares (approved):

semgrep_output_v1.atd
aaronmichaelacosta: AFAIK npm/yarn and pip/poetry are cases in which two package managers use the same _central repository_. So, for example, a package with the same name/version in pip is guaranteed to be the exact same as the package with the same name/version in poetry.

While Cocoapods and SwiftPM have some coincidental overlap (e.g. packages can live in github), but they are completely separate ecosystems with different registries. There's no guarantee that a package in one will exist in the other, and it's entirely possible for two packages to exist in the PMs with the same name/version but be completely different.

PR #332, opened 2025-01-09
chore: add `id_info` to `ImportFrom` 

brandonspark:
As title.

It seems that our `Makefile` is improperly specified and does not regenerate `ast_generic_v1.py`. I fixed this in this PR also.

	  generated by Semgrep 1.50.0.


aryx (approved):
There is no need really for those changes. Nobody is using
ast_generic_v1.atd and using the generic AST from other languages.
We should probably remove the file actually.

PR #333, opened 2025-01-09
Remove semgrep_version from project_metadata and meta from scan_request

aryx:
The backend should not use those fields anymore so
we can remove them. Note that even if old versions
of semgrep are still producing those fields, it is fine
because when the backend is consuming the JSON, any unknown
field is skipped and those fields should not be used anymore.

test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


aryx:
Followup to https://github.com/semgrep/semgrep-proprietary/pull/1965

mjambon (approved):

chmccreery (approved):

chmccreery:
> @chmccreery you'll have to use this version and double check that it compiles with semgrep-app frontend and backend before we merge this.

Done! See https://github.com/semgrep/semgrep-app/pull/18439

aryx:
Ok @chmccreery let's wait that nat is releasing 1.103.0 and then we can merge this and the semgrep and semgrep-app PRs.

aryx:
It's merged! @chmccreery you can merge your PR in semgrep-app!

PR #335, opened 2025-01-10
chore(sca): add flag to disable subproject change detection

bkettle:
Adds a new flag to pass from the app to the CLI that forces the CLI to resolve
dependencies for all subprojects, all the time (even for diff scans). Some
customers want this so that we get a full picture of their dependencies for
every branch.

	  generated by Semgrep 1.50.0.


salolivares (approved):

semgrep_output_v1.atd
```diff
@@ -1153,6 +1153,7 @@ type features = {
    ~deepsemgrep: bool;
    ~dependency_query: bool;
    ~path_to_transitivity: bool;
+   ~scan_all_deps_in_diff_scan: bool;
```
aryx, line 1156: can be useful to add a comment here explaining quickly what the feature is.

aryx (approved):
You need to rebase/merge on develop

PR #336, opened 2025-01-13
simplify the RPC for SARIF

aryx:
The goal is to make it as close as possible as the general
CallFormat so at some point we can merge them.

test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


mjambon (approved):

aryx:
The lint checks are failure are normal. We do not care about backward compatibility for RPCs types.

PR #337, opened 2025-01-14
New input type for semgrep-core allows taking scanning roots instead of target files

mjambon:
This is used by https://github.com/semgrep/semgrep-proprietary/pull/2878.

	  generated by Semgrep 1.50.0.


aryx:
BTW would be great after all this work to merge the semgrep-interfaces repo in semgrep-pro :) Those double PRs
are becoming annoying (just like double PRs we used to have with change in semgrep-oss and semgrep-pro).

mjambon:
This is annoying:
```
+[0db7c783] Incompatibility in both directions:
+Incompatible kinds of types: list/array is now a sum type or enum.
+The following types are affected:
+  targets
```

It's a legitimate incompatibility report but it's about the semgrep-core interface that we can break at will. It wouldn't happen if we could split the ATD file into multiple modules each concerned with a different interface. Maybe for now we could have some kind of hack to work around this. Atddiff offers an option to only check certain types so we could use this but it's not great for the long term (since adding a new interface would require adding the new root types to the list of types to check). By increasing difficulty (and cleanliness), we have:
1. use some shell script hack to silence reports about semgrep-core interface changes
2. use atddiff with the `--types` option to only check the root types involved in semgrep's external interfaces (= not semgrep-core)
3. modify atddiff to ensure any root type is listed on the command line by either `--types` or `--ignore-types` so as to ensure we don't forget to check new types we should check
4. implement and ship [atdml with module support](https://github.com/ahrefs/atd/issues/265)

aryx:
Yes I've asked a few times for the ability to white list a few roots to not bother to check anything that descend from RPC or semgrep-core stuff.

aryx:
whitelist or blacklist. In this case we probably want atddiff --do-not-bother-about-those-roots 'function_call,function_return,core_output'

Note that core_output also deeply use cli_output_extra, but this cli_output_extra is also used by cli_output so we should
run the compatibility check on cli_output_extra (but not on types that are referenced only by core_output).

mjambon:
> whitelist or blacklist. In this case we probably want atddiff --do-not-bother-about-those-roots 'function_call,function_return,core_output'
>
> Note that core_output also deeply use cli_output_extra, but this cli_output_extra is also used by cli_output so we should
run the compatibility check on cli_output_extra (but not on types that are referenced only by core_output).

I added an option to atddiff to ensure that we don't miss future type definitions. atd PR coming soon.

mjambon:
@aryx you wrote at https://github.com/ahrefs/atd/pull/422#issuecomment-2602016929:
> the list for --types and --ignore seems wrong.
> I don't see cli_output for example, which should be in --types I guess, and
> function_call and function_return should be in --ignore as they don't have to be backward compatible.

`cli_output` is used by `function_call` and the latter wasn't in `--types`. I'll make this change.

PR #340, opened 2025-01-17
Remove ast_generic_v1.atd

aryx:
It was experimental and it's not maintained nor used, so
let's get rid of it.

test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


yosefAlsuhaibani:
Looks good, Thank you!

yosefAlsuhaibani (approved):

nmote:
SO EXCITED

mjambon (approved):

PR #341, opened 2025-02-04
feat(scrat): symbol analysis

brandonspark:
This PR adds "symbol analysis", a dump of the third-party symbols used by a project, to the scan results for Semgrep.

This will enable the `scrat` project.

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -2167,6 +2169,37 @@ type resolution_result = [
...
+
+   We may be interested in this for the purposes of analyzing the vulnerability
+   of the repo based on its dependencies, and providing more customized/informed
+   guidance on how to remediate SCA vulns.
+
+   For instance, source code like
+   ```
+   from werkzeug.wrappers import Request
+   ```
+   may be described as using the "symbol" [werkzeug, wrappers, Request].
```
aryx, line 2187: There is some PRIVACY issue here; we're not sending the code of the customer but the symbols/libs it used
so be cautious here.

semgrep_output_v1.atd
```diff
@@ -2167,6 +2169,37 @@ type resolution_result = [
...
+ *)
+
+(* A symbol is a FQN. *)
+type symbol = string list
```
aryx, line 2191: I would make it a record so it can be changed later in a backward compatible way (ATD has good support
for adding fields in a record and be backward compatible; not so much for other types).
type symbol = {
  fully_qualified_name: string list;
}

semgrep_output_v1.atd
```diff
@@ -2167,6 +2169,37 @@ type resolution_result = [
...
+
+(* We store the location of the usage, because we may want to be able to know how
+   many uses of the symbol there are, and where.
+ *)
+type symbol_usage = {
+  symbol: symbol;
+  loc: location;
```
aryx, line 2198: could be a locs: location list; here
to avoid repeating the same symbol again and again (which could save some space).

semgrep_output_v1.atd
aryx: or maybe fqn: string list;
to save some space in the sent JSON.

aryx (approved):
I think ultimately we want to move closer to SCIP.proto
but let's go forward with this to unblock you.

PR #346, opened 2025-02-07
pysemgrep upload symbol analysis

brandonspark:
This adds some stuff which makes us able to call uploading symbol analysis from pysemgrep.
 
	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -2330,6 +2339,7 @@ type function_return <python decorator="dataclass(frozen=True)"> = [
   | RetSarifFormat of string
   | RetValidate of bool
   | RetResolveDependencies of (dependency_source * resolution_result) list
+  | RetUploadSymbolAnalysis of string
```
aryx, line 2342: what is the string? Error message?

aryx (approved):

aryx:
This data and new endpoint is potentially a big security risk though. We are now collecting the symbols in the
customer's code, and if this go out this is very bad because people could know a lot about the code of the customer, the kind of library they use etc.
cc @werrett

PR #347, opened 2025-02-10
Add GetTargets semgrep-core RPC

nmote:
	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -1990,7 +1998,6 @@ type targeting_conf <ocaml attr="deriving show"> = {
   force_novcs_project : bool;
   exclude_minified_files : bool;
   ?baseline_commit : string option;
-  diff_depth : int;
```
emjin, line 1993: Is there a reason to remove `diff_depth` in this PR? I would generally prefer to separate concerns

emjin (approved):

semgrep_output_v1.atd
nmote: I'm not sure, to be honest. I considered separating it out but given the minor annoyance in updating the submodule I left it in. I'll split it though.

mjambon: This is a mechanical change needed to mirror the other targeting_conf type we have in OCaml where Pad removed the `diff_depth` field. I wouldn't bother with another PR since this is unused. Also, this is an internal interface which can change without worrying about backward compatibility.

PR #350, opened 2025-02-12
Define new sca_error intermediate type and sca_resolution_error and ScaParseError

aryx:
test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -768,7 +768,9 @@ type error_type
   | PatternParseError0 <json name="Pattern parse error">
   | IncompatibleRule0 <json name="Incompatible rule">
   (* since semgrep 1.94.0 *)
-  | DependencyResolutionError of resolution_error
+  | DependencyResolutionError of resolution_error_kind
+  (* since semgrep 1.109.0 (to replace dependency_parser_error) *)
+  | ScaParseError of sca_parser_name
```
bkettle, line 773: We could alternatively skip adding an extra variant here, and instead add a variant to `resolution_error_kind`. In my head "dependency resolution" refers to any way that we determine the dependencies for a project, so this could make sense. What do you think?

aryx: Sounds good to me. That would maybe the need to have all those Union[out.DependencyParseError, out.ScaResolutionError] ? and we would have just out.ScaResolutionError or maybe just out.ScaError

aryx: What would be the difference between this ScaParserError of sca_parser_name and the
ParseDependenciesFailed of string in resolution_error_kind?

bkettle: `ParseDependenciesFailed` was originally intended to be for when we produce some dependency list in lockfileless scanning (by talking to the package manager) but fail to parse it correctly, while `ScaParserError` is that a lockfile parser failed

PR #353, opened 2025-02-19
Define TR caching comms in ATD

aryx:
To make things more concrete about the interface we want
with the backend.

test plan:
make

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -1831,6 +1833,62 @@ type scan_config = {
...
+ * URL should hopefully not change (we could md5sum it just in case) and
+ * the content of the rule_id should also not change (could md5sum it maybe too).
+ * I've added tr_version below just in case we want to invalidate past
+ * cached entries (e.g., the semgrep engine itself changed enough that
+ * some past cached results might be wrong and should be recomputed)
+*)
+type tr_cache_key = {
+    rule_id: rule_id;
+    (* ex: http://some-website/hello-world.0.1.2.tgz like in found_dependency *)
+    resolved_url: string;
```
bkettle, line 1857: I don't know if this will be easy to find in many cases, though I agree that if it's possible it makes the best key. I guess it is probably fine to start with this, and add another key later if needed.

semgrep_output_v1.atd
```diff
@@ -1831,6 +1833,62 @@ type scan_config = {
...
+(* The "value" *)
+type tr_cache_match_result = {
+    (* TODO: could define separate sca_transitive_match type? Should only be
+     * one of TransitiveXxx case of sca_match_kind
+     * TODO? add other fields from sca_match or just sca_match_kind is enough?
+     * add location maybe? or make it part of the [transitive_reachable]
+     * and [transitive_unreachable] records?
+     * TODO? make it a list? match_results: ... list; ?
+     *)
+    match_result: sca_match_kind;
```
bkettle, line 1873: This is a bit odd, since when scanning a package with a rule it will result in direct code matches, not sca matches. Maybe we should just return the match here? Then, the CLI could treat those matches the same as matches that it receives from a call to Semgrep locally.

semgrep_output_v1.atd
aryx: yes we can always refine. This is just defining the interface. Once we start the implementation we will discover
we need to refine it.

semgrep_output_v1.atd
aryx: We could. This is a bigger data structure to store then though, and for TR what we really need is actually just the sca_transitive_match_kind; that's the thing we try to optimize to avoid downloading the dependency and run semgrep on it.

bkettle: I think we should at least store the match locations. I don't think it makes sense to store `sca_match_kind` because if there are matches in multiple packages, we will somehow need to combine those into a single finding in the cli

bkettle (approved):
lgtm

gautambhat (approved):

PR #354, opened 2025-02-20
Define Subproject datastructure in ATD (from subproject.py)

aryx:
test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -2289,13 +2297,78 @@ type sca_error = [
...
+] <ocaml repr="classic">
+
+type unresolved_subproject
+    <python decorator="dataclass(frozen=True)"> =
+{
+   info: subproject;
+   reason: unresolved_reason;
+   (* this is set only when the reason is UnresolvedFailed
+    * TODO: add it in unresolved_reason instead?
+    *)
```
bkettle, line 2370: it is theoretically possible to have partial failures, which was my original thinking for having this be separate (e.g. for parsers we can have errors but still have some result)

PR #355, opened 2025-02-24
feat: expose metavariable-name.modules

kopecs:
Allows metavariable-name to appear without a semgrep-internal prefix when using the modules constraint (but not yet for kind).

	  generated by Semgrep 1.50.0.


kopecs:
This might needs some minor tweaks to ensure error messages are still OK.

inkz (approved):

PR #356, opened 2025-02-25
feat(sca): add unresolved reason and errors to subproject stats

salolivares:
In https://github.com/semgrep/semgrep-proprietary/pull/2872, we updated the SCA subproject logic to scan only changed subprojects in diff scans. Subprojects that remained unchanged were intentionally left unresolved with the reason UnresolvedSkipped.

Previously, we didn't include this information in our subproject stats, leaving us without visibility into why a subproject was unresolved. This PR introduces two optional fields, unresolved_reason and errors, allowing us to capture more granular stats and better understand subproject resolution failures.

	  generated by Semgrep 1.50.0.


bkettle (approved):
this will be great, thanks

semgrep_output_v1.atd
```diff
@@ -1730,6 +1730,10 @@ type subproject_stats = {
   dependency_sources: dependency_source_file list;
   (* Results of dependency resolution, empty if resolution failed *)
   ?resolved_stats: dependency_resolution_stats option;
+  (* Reason why resolution failed, empty if resolution succeeded *)
+  ?unresolved_reason: unresolved_reason option;
+  (* Errors encountered during subproject resolution *)
+  ?errors: sca_error list option;
```
aryx, line 1736: for list you can also use ~errors: sca_error list;
and it's backward compatible. It will just create an empty list by default when you look at the generated code.
The ocaml code (and probably Python code) is a little bit more convenient to use when you use ~.
@mjambon can maybe add more explanations about ~ but otherwise there is the ATD ref manual online.

salolivares: oh nice. i'll switch to that.!

salolivares: done in [688c5e7](https://github.com/semgrep/semgrep-interfaces/pull/356/commits/688c5e769de68e49e4f4bba60013e0011ca05ddf)

PR #358, opened 2025-02-27
Remove depth_diff field that's irrelevant to target selection

mjambon:
This is for https://github.com/semgrep/semgrep-proprietary/pull/3099

	  generated by Semgrep 1.50.0.


nmote (approved):
The change here seems fine based on your previous explanations of it, but before merging could you update the PR title to match the change, and explain the rationale behind the removal of `diff_depth` here?

mjambon:
Somehow, the relevant changes were merged already, some rebasing happened, and now we're left with only this unrelated change.

PR #359, opened 2025-02-28
chore(sca): tweak unresolved version

bkettle:
Add "disabled" option to `UnresolvedReason` and make it a frozen dataclass so we
can use it in sets and as a dictionary key in python

	  generated by Semgrep 1.50.0.


aryx (approved):
maybe add some comments for each case, but espcially for Disabled

PR #361, opened 2025-03-05
chore: use opam to simplify makefile

ajbt200128:
Let's install check-jsonschema via depexts!
See https://github.com/semgrep/semgrep-proprietary/pull/3273 for more context

	  generated by Semgrep 1.50.0.


mjambon (approved):

semgrep-interfaces.opam
```diff
@@ -0,0 +1,31 @@
...
+
+#TODO: we don't build on ubuntu or debian ever, and they are not packages there,
+# so let's just ignore it for now
+#["check-jsonschema"] {os-family = "debian"}
+#["check-jsonschema"] {os-family = "ubuntu"}
+depexts: [
+   ["check-jsonschema"] {os-distribution = "alpine"}
+   ["check-jsonschema"] {os = "macos" & os-distribution = "homebrew"}
+   ["check-jsonschema"] {os = "macos" & os-distribution = "macports"}
+   ["check-jsonschema"] {os = "win32" & os-distribution = "macports"}
```
aryx, line 30: there is macports on win32??

PR #362, opened 2025-03-11
Replace lockfile with dependency source in targets

mmcqd:
	  generated by Semgrep 1.50.0.


mmcqd:
I'm confused by the proto conflict, not sure what's causing that

aryx:
just pull back the latest main, rebase, and rerun make to regenerate everything

semgrep_output_v1.atd
```diff
@@ -2023,7 +2023,7 @@ type analyzer <ocaml attr="deriving show"> = string wrap <ocaml module="Analyzer
 *)
 type target <ocaml attr="deriving show"> = [
```
aryx, line 2024: We could also change the targets type and have there a list of code_target and a list of dependency_source.
But let's go with this approach.
Note that @mjambon added also a new Scanning_roots cases in the targets type where pysemgrep relies on
osemgrep file targeting to compute the targets, but I think we're fine with your change; just in the futur for those
cases where pass the Scanning_roots, it will also be the job of the OCaml code to compute the dependency_source (using @bkettle subproject matcher probably).

aryx (approved):

PR #366, opened 2025-03-20
Add some <ocaml repr="classic"> to a few SCA types

aryx:
I don't like polymorphic variants (`Xxx) and prefer classic variants.

Also rename transitivity to dependency_kind

test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -516,28 +516,30 @@ type ecosystem
   | Mix <json name="mix">
   | Hex <json name="hex">
   | Opam <json name="opam">
-]
+] <ocaml repr="classic">
 
-type transitivity
+type dependency_kind
```
bkettle, line 521: I am a little hesitant about this change, I think we might introduce additional categories like dev dependencies that might make this name confusing. But I guess this is changeable in the future so not a big deal

semgrep_output_v1.atd
```diff
@@ -495,7 +495,7 @@ type match_intermediate_var <python decorator="dataclass(frozen=True)"> = {
 (* both ecosystem and transitivity below have frozen=True so the generated
  * classes can be hashed and put in sets (see calls to reachable_deps.add()
  * in semgrep SCA code)
- * TODO: use <ocaml repr="classic">, and do the same for manifest
+ * alt: type package_manager
```
bkettle, line 498: I would prefer "namespace" over package manager, since e.g. both maven and gradle package managers use the maven ecosystem, since they pull packages from the same place

bkettle (approved):

semgrep_output_v1.atd
aryx: I agree. I hesitated too, but seeing some out.Transitivity(out.Transitive) convinced me :)

semgrep_output_v1.atd
aryx: yes but namespace is even more vague.

PR #368, opened 2025-03-26
chore(tr): allow multiple dependency source paths

bkettle:
Previously, TR assumed that all source code for a dependency was contained in a single folder. However, pip unpacks packages into potentially multiple folders. To support `uv` and other python package managers, we need to allow specifying source files directly.

	  generated by Semgrep 1.50.0.


aryx (approved):

aryx:
Not having a single dir fpath will be annoying I think.
I'm not convinced we should allows multiple Fpath.t ...
if it's just for this setuptools package.
I would vote for a single dir for now to make our life simpler.

aryx:
If we start to get the parents of 2 folders, we will end up with node_modules/   or site-packages/
and the file targeting in semgrep might just analyze all the modules ...

aryx:
For packages that unpack in 2 folders I would just Logs.warn and pick one (and preferably the one that is close
to the package name).

bkettle:
> Not having a single dir fpath will be annoying I think. I'm not convinced we should allows multiple Fpath.t ... if it's just for this setuptools package. I would vote for a single dir for now to make our life simpler.

I don't think setuptools is the only example, it was just the first one I could find. Is it true that file targeting will look at everything under the scanning root, even if we pass explicit targets?

PR #371, opened 2025-04-07
Improve explanations for SCA TR findings

aryx:
test plan:
see related PR in semgrep

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -606,14 +603,30 @@ type sca_match_kind = [
...
-  explanation: string option;
+  (* The matches we found in 3rd party libraries.
+   * Ideally the location in cli_match are relative to the root of the project
+   * so one can display matches as package@/path/to/finding.py
+   * TODO: include matched code? 3rd party libraries are usually OSS
+   * so maybe ok to store code for once in our DBs? Otherwise The App will
+   * have to redownload and reaccess the package code (or maybe we
+   * could give a github URL?)
+   *)
+  matches: (string (* package uri? *) * cli_match list) list;
```
bkettle, line 614: I think `cli_match` includes `found_dependency`, we may not need the string here

semgrep_output_v1.atd
```diff
@@ -606,14 +603,30 @@ type sca_match_kind = [
...
+  (* some extra explanation that the user can understand *)
+  explanation: string option;
 }
 
 type transitive_unreachable = {
-  (* TODO: some proof that the user can understand *)
+  (* We didn't find any findings in all the 3rd party libraries that are using
+   * the 3rd party vulnerable library. This is a "proof of work".
+   *)
+  analyzed_packages: string (* package uri? *) list;
```
bkettle, line 628: I'm iffy on using just a string to identify packages. It's going to be hard to find the package name and version from that, which is what we probably want to display to the user. Maybe we should use `found_dependency` here? Or maybe just name/version/ecosystem triplet?

bkettle (approved):

semgrep_output_v1.atd
```diff
@@ -594,6 +587,10 @@ type sca_match_kind = [
    * of the SCA rule in 3rd party code, or if there is
    * it's in third-party code that is not accessed from the
    * 1st-party code (e.g., via callgraph analysis)
+   * Note that there is no need for DirectUnreachable because semgrep
+   * would never generate such a finding. We have TransitiveUnreachable
+   * because semgrep first generates some Undetermined that we then
+   * retag as DirectUnreachable.
```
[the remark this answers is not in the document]
aryx: Yes, TransitiveUndetermined

semgrep_output_v1.atd
aryx: hmm but when we run semgrep on third party code we remove the SCA part, so we should not get found_dependency.

aryx: but yes I could use `(found_dependency * cli_match list) list`

aryx:
The linting errors are normal; we are indeed incompatible with the previous version but that's ok because no customer generated
the old data because TR is not enabled yes so we can still experiment and change the data structures.

PR #372, opened 2025-04-16
Add semgrepignore_filename option

mjambon:
This is to support the new `--semgrepignore-filename` option: https://github.com/semgrep/semgrep-proprietary/pull/3680

	  generated by Semgrep 1.50.0.


aryx (approved):
We really want to make this configurable? Who needs that?

mjambon:
> We really want to make this configurable? Who needs that?

We'd be the main users. This for scanning test targets without having to copy them into their own project/repo. The alternative we've been using is to set `--project-root` but it doesn't work with `semgrep ci`.

We could mark this option as "experimental" if we don't want to commit to maintaining it. Yeah, I'll do that.

mjambon:
I renamed the command-line option to `--x-semgrepignore-filename` and documented it as internal (in https://github.com/semgrep/semgrep-proprietary/pull/3680).

PR #373, opened 2025-04-16
Update some language extensions based on existing test files

nhakmiller:
### Changes

I observed some test cases from https://github.com/semgrep/semgrep-rules that were not being tagged with a language while being loaded because they did not have their extensions listed here.

I added the extensions I observed for the languages they belong to. I'm mostly tracking the impact this will have on the registry loading & running tests, but if there are severe other implications of this change I can retract it.

### Testing
I did not edit a `.atd` file, so I did not run make.
 
	  generated by Semgrep 1.50.0.

### DIFF
--- a/lang.json
+++ b/lang.json
@@ -151,7 +151,8 @@
     ],
     "maturity": "ga",
     "exts": [
-      ".cs"
+      ".cs",
+      ".cshtml"
     ],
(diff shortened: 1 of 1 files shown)


aryx (changes requested):
old

PR #375, opened 2025-05-09
Keep track of ppaths after target discovery so as to allow rule filtering

mjambon:
The ppath of a target file is its path relative to the project root identified on the OCaml side during target discovery (`Find_targets` module). It is now returned to pysemgrep so that rule filtering based on the paths.include/exclude patterns can take place correctly (as per the Semgrepignore/Gitignore spec) on the ppaths in Python and then again in OCaml when the targets are sent to semgrep-core.

Goes with https://github.com/semgrep/semgrep-proprietary/pull/3859

	  generated by Semgrep 1.50.0.


nmote (approved):

ajbt200128 (approved):

mjambon:
https://github.com/semgrep/semgrep-proprietary/pull/3859 is (finally) ready. Merging!

PR #379, opened 2025-05-28
feat: add subproject resolution fields to CLI output schema

bkettle:
Adds information about subprojects and dependency resolution to the scan JSON output. This is intended to be available only for logged-in users.

	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -1050,6 +1050,11 @@ type cli_output_extra = {
 
     (* EXPERIMENTAL: since: 1.37.0 *)
     ~skipped_rules: skipped_rule list;
+
+    (* SCA subproject resolution results 
+     * EXPERIMENTAL: since: 1.109.0 *)
```
bkettle, line 1055:
```suggestion
     * EXPERIMENTAL: since: 1.123.0 *)
```

semgrep_output_v1.atd
```diff
@@ -1050,6 +1050,10 @@ type cli_output_extra = {
 
     (* EXPERIMENTAL: since: 1.37.0 *)
     ~skipped_rules: skipped_rule list;
+
+    (* SCA subproject resolution results 
+     * EXPERIMENTAL: since: 1.123.0 *)
```
TikhonJelvis, line 1055: is this comment just a convention, or do we have some tooling that can parse these annotations?

TikhonJelvis (approved):
makes sense

semgrep_output_v1.atd
bkettle: just convention

PR #384, opened 2025-06-06
output: time: Add quick profiling 'parsing_time' data

IagoAbal:
	  generated by Semgrep 1.50.0.


semgrep_output_v1.atd
```diff
@@ -978,13 +978,31 @@ type profile = {
...
     ?max_memory_bytes : int option;
   }
 
+type file_time <ocaml attr="deriving show"> = {
+    fpath: fpath;
+    ftime: float;
+}
+
+type stats = {
+    mean: float;
```
IagoAbal, line 996: We could also have a `total: float` field here. I'd be fine with that, it's just the field name `per_file_time` makes it more clear what the data is, and with that name it felt a bit stranger to include the total.

emjin:
While we're making this change, should we also add some fields to record slowest rules?

semgrep_output_v1.atd
emjin: Maybe we could rename the field to "summary_stats"?

PR #393, opened 2025-07-29
output(validation): Pass actual errors instead of status flag only

punchagan:
	  generated by Semgrep 1.50.0.


emjin (approved):
lgtm

PR #396, opened 2025-08-14
feat: add metrics for mcp rpc calls to semgrep

liukatkat:
- [x] I ran `make setup && make`to update the generated code after editing a `.atd`file (TODO: have a CI check)
For example, the Semgrep backend need to still be able to _consume_data
generated by Semgrep 1.50.0.
- [x] Any accompanying changes in `semgrep-proprietary`are approved and ready to merge once this PR is merged

### DIFF
--- a/semgrep_metrics.atd
+++ b/semgrep_metrics.atd
@@ -270,7 +270,6 @@ type extension = {
 }
 
 type mcp = {
-    ?deployment_id <ocaml mutable>: int option;
     ?session_id <ocaml mutable>: string option;
     ?num_findings <ocaml mutable>: int option;
 }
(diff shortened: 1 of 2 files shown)


brandonspark (approved):
Generally, I am OK with this. My only thought though is that maybe some point in the future, we will want to change `num_findings` to more detailed information about the findings (which rule fired, how many findings per rule, which products did each finding correspond to).

In which case, is this change backwards-compatible, from the metrics consumption side? Or are we doomed to always have to process fields we add here? Maybe @ajbt200128 will know.

liukatkat:
but  i think it should only break my dashboard and shouldn't affect anything else because i am just extending the json blob.

liukatkat:
### Merge activity

* **Sep 2, 11:09 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/semgrep/semgrep-interfaces/396).

PR #399, opened 2025-08-15
feat: add a new rule property for custom entropy values

emjin:
Corresponding PR: https://github.com/semgrep/semgrep-proprietary/pull/4386

	  generated by Semgrep 1.50.0.


jmgrosen:
Closing both because this seems stale but also because semgrep-proprietary is now the source of truth for semgrep-interfaces.

PR #400, opened 2025-08-19
chore(cli): revert unneeded new parameter

nhakmiller:
### Description

Reverting https://github.com/semgrep/semgrep-interfaces/pull/397 because we have a better solution that does not need this parameter.

I talked with @salolivares and we think this might be fine even though it's not backwards compatible because neither the server nor the CLI ever actually started using this parameter. But I'm actually not 100% on if it's ok to make a non backwards compatible change even if the parameter was never set or read by anyone so I'd be interested to hear from someone who could confirm that.

	  generated by Semgrep 1.50.0.


nhakmiller:
https://github.com/semgrep/semgrep-proprietary/pull/4481 is approved, so going to merge this now then that one.

PR #401, opened 2025-08-21
feat(contributors): Support populating scan metadata to results for contributor processing

zzeleznick:
# Description

To support processing contributors, we want to associate each contribution with the driving scan for billing. 

	  generated by Semgrep 1.50.0.
- [N/A] Any accompanying changes in `semgrep-proprietary` are approved and ready to merge once this PR is merged


semgrep_output_v1.atd
```diff
@@ -1745,6 +1745,23 @@ type ci_scan_results = {
...
+}
+
+(* Scan metadata populated by the backend after receiving
+ * the scan results from the CLI via POST request to
+ * /scans/<int:scan_id>/results
+ *)
+type ci_scan_metadata = {
+  scan_id: int;
+  deployment_id: int;
+  repository_id: int;
```
tpetr, line 1761: `repository_ref_id` might be useful too (if not for contribs, def for other users of this data)

semgrep_output_v1.atd
```diff
@@ -1745,6 +1745,23 @@ type ci_scan_results = {
    (* this data was originally sent to /complete, but we want to start sending
     * it to /results *)
    ?dependencies: ci_scan_dependencies option;
+
+   (* filled in by the backend to associate contributions with their corresponding
+    * source repository and driving scan*)
```
tpetr, line 1750: uber-nit but there will probably be other consumers of this data, so maybe we just stick to the what instead of the why (or change the wording to "for instance ...")

tpetr (commented):
LGTM but might want to get an additional blessing from scan pipeline owners (or atleast a heads up) (cc @semgrep/sms)

amchiclet (approved):
I think I'm generally OK with this and added some comments. Thanks!

semgrep_output_v1.atd
```diff
@@ -1745,6 +1745,23 @@ type ci_scan_results = {
...
+
+   (* filled in by the backend to associate scan results with the driving scan*)
+   ?metadata: ci_scan_metadata option;
+}
+
+(* Scan metadata populated by the backend after receiving
+ * the scan results from the CLI via POST request to
+ * /scans/<int:scan_id>/results
+ *)
+type ci_scan_metadata = {
```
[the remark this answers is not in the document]
zzeleznick: can do – not 100% sure on the repo ref id but believe it might help associate repos with their branches

PR #403, opened 2025-08-26
chore(sca): make dependency types use frozen dataclasses"

bkettle:
Updates the types for dependency resolution results to use frozen dataclasses.
This allows us to create sets from them, making it easier to compare in tests.

	  generated by Semgrep 1.50.0.


bkettle:
`allowed_hashes` is a dict and isn't super trivial to change (lots of usages) so this isn't especially useful, closing

PR #405, opened 2025-09-05
feat: only allow valid lang keys in the editor

milanwilliams:
	  generated by Semgrep 1.50.0.
- [X] Any accompanying changes in `semgrep-proprietary` are approved and ready to merge once this PR is merged

SAF-1953


emjin:
@milanwilliams The backwards compatibility summary works for the check! You should make a PR in semgrep-proprietary to point its `semgrep-interfaces` submodule to develop (just `cd semgrep-interfaces` and `git checkout develop`, then go back to the root and you should see that `semgrep-interfaces` was modified)

yosefAlsuhaibani (approved):
This seems fine! thank you 🙇

punchagan:
@yosefAlsuhaibani What is the recommended way to keep this list of languages updated? I see that `xml` was added in #409 and I added `gosu` in #408. Is it manually added? Would it be worth adding a test to catch missing languages based on what's generated in `generate.py`, if not getting it to update the `rule_schema_v1.yaml` file?

PR #406, opened 2025-09-10
chore: Add new Manifest.kind for build.gradle.kts files

TikhonJelvis:
	  generated by Semgrep 1.50.0.


bkettle (approved):
We'll need to update the subproject matchers to output these as well, plus anywhere that BuildGradle is mentioned in the code

TikhonJelvis:
Yep, I just pushed a couple of stack PRs that (I *think*) update all the code that needs updating.

PR #445, opened 2026-03-25
feat(supply-chain): add support for version ranges to dependency definitions

nhakmiller:
### Context

I'm investigating what it could look like to support unpinned versions of dependencies, at least in processing `requirements.txt` (a special case of dependency source file which is kind of like a manifest file but also kind of like a lockfile.

This change adds a field for us to optionally track a `version_specifier` on a `found_dependency` in addition to a package name and version. We will consume this on the CLI to track non exact pinned versions.

- [x] I ran `make setup && make` to update the generated code after editing a `.atd` file
	  generated by Semgrep 1.50.0.


bkettle (commented):
the other thing we need to think about is how the app will handle this stuff and put it into its database, since the DB currently expects an exact version as well. You probably know the most about what's feasible there from your experience with the version filters

nhakmiller:
> the other thing we need to think about is how the app will handle this stuff and put it into its database, since the DB currently expects an exact version as well. You probably know the most about what's feasible there from your experience with the version filters

As of right now, app don't care.

In the findings page:
Screenshot 2026-03-30 at 2 48 06 PM

In the dependencies page:
Screenshot 2026-03-30 at 2 47 29 PM

It works'ish as is with dependency version filters, for reasons that are not clear to me:
Screenshot 2026-03-30 at 2 49 33 PM
Screenshot 2026-03-30 at 2 50 06 PM

That being said, I think the filtering is going to be unintuitive no matter how it's arranged. Especially for things like exact matches. We might just need to say that basically filtering on versions is best effort, but I think that will not go overwell. Perhaps we can extend the version filtering to support these values, but we probably want to do an exercise of mapping the expected behaviors for these filters.

